PR #267, opened 2014-11-28
Default configuration

ababol:
Let's talk about the new default configuration.

Should we keep the old one?

I really like the new inline view but the status bar becomes a little bit useless if the inline view is displayed ; except if you `Show all errors` in the status bar, that's why I put this configuration by default in the commit.


dmnd:
agreed, sorry for the delay on this. merging now.

ababol:
Nice,  thanks!

PR #290, opened 2014-12-15
Add public method beforeSpawnProcess: (command, args, options) to Linter

### LINKED ISSUE #289: Feature Request: modify command and args before create new BufferedProcess
I'm working on Linter Rust. Currently we use `rustc` (Rust's compiler) as the executablePath to lint Rust sources, this works, but have problem: you can't just compile the only one source file of a large project which have many source files. Rust already have a package manager, Cargo, which is largely used to build Rust projects and their dependencies.

My plan is, make the Linter Rust to support `rustc` and `cargo` at the same time: if `Cargo.toml` is found, use `cargo`, or else use `rustc`.

But while I working on this, I found that Linter itself has problems when create new BufferedProcess:
- It always use executblePath (`rustc` in Linter Rust) as the *command*
- It always add the source file being editing as the last arg of *args*

I can't change these behaviors inside Linter Rust. But I do need modify the command (either `rustc` or `cargo`) and args (`Cargo.toml` instead of the file user is editing), to improve Linter Rust. I think we could modify Linter to support this. If a linter have the chance to change the command and args before create new BufferedProcess, that would be OK.

Thank you!

liigo:
Gives subclasses a chance to read or change the command, args and options, before creating new BufferedProcess while lintFile.

This is a backward compatible improvement.

Closes #289

PS: I'a a newbie of CoffeeScript .

### DIFF
--- a/lib/linter.coffee
+++ b/lib/linter.coffee
@@ -149,6 +149,9 @@ class Linter
         else data = dataStderr
       @processMessage data, callback
 
+    {command, args, options} = @beforeSpawnProcess(command, args, options)
+    log("beforeSpawnProcess:", command, args, options)
+
     process = new BufferedProcess({command, args, options,
                                   stdout, stderr, exit})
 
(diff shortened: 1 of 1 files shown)


jhasse:
Any news on getting this merged?

KingOfThePirates:
I would like news too :)

iam4x:
Let's merge it, though @dmnd was on it since he answered first :)

PR #277, opened 2014-12-03
Status bar summary

ababol:
screen shot 2014-12-03 at 21 46 54

So it display the number of errors/warning.

Let me know if you need further information about the commit.

See: https://github.com/AtomLinter/Linter/issues/242

PR #309, opened 2014-12-22
Add support for message level 'info'.

beck:
When styling markers and a line has multiple messages, choose the message
with the level of highest precedence.

Default view:
screen shot 2015-01-21 at 00 28 17

ShowInfoMessages turned on: 
screen shot 2015-01-21 at 00 27 30

### DIFF
--- a/lib/linter-view.coffee
+++ b/lib/linter-view.coffee
@@ -89,9 +89,9 @@ class LinterView
         @showHighlighting = showHighlighting
         @display()
 
-    @subscriptions.add atom.config.observe 'linter.hushInfoMessages',
-      (hushInfoMessages) =>
-        @hushInfoMessages = hushInfoMessages
+    @subscriptions.add atom.config.observe 'linter.showInfoMessages',
+      (showInfoMessages) =>
+        @showInfoMessages = showInfoMessages
         @display()
 
   # Internal: register handlers for editor buffer events
@@ -177,11 +177,11 @@ class LinterView
 
   # Internal: Pidgeonhole messages onto lines. Each line gets only one message,
   # the message with the highest level presides. Messages of unrecognizable
-  # level (or hushed by config) will be skipped.
+  # level (or silenced by config) will be skipped.
   sortMessagesByLine: (messages) ->
     lines = {}
-    levels = ['info', 'warning', 'error']
-    levels.shift() if @hushInfoMessages
+    levels = ['warning', 'error']
+    levels.unshift('info') if @showInfoMessages
     for message in messages
       lNum = message.line
       line = lines[lNum] || { 'level': -1 }
--- a/lib/linter-view.coffee
+++ b/lib/linter-view.coffee
@@ -147,13 +147,12 @@ class LinterView
   processMessage: (messages, tempFileInfo, linter) =>
     log "#{linter.linterName} returned", linter, messages
 
+    @messages = @messages.concat(messages)
     tempFileInfo.completedLinters++
     if tempFileInfo.completedLinters == @linters.length
+      @display @messages
       rimraf tempFileInfo.path, (err) ->
         throw err if err?
-
-    @messages = @messages.concat(messages)
-    @display()
 
   # Internal: Destroy all markers (and associated decorations)
   destroyMarkers: ->
--- a/lib/linter-view.coffee
+++ b/lib/linter-view.coffee
@@ -1,4 +1,4 @@
-_ = require 'lodash'
+lodash = require 'lodash'
 fs = require 'fs'
 temp = require 'temp'
 path = require 'path'
@@ -45,8 +45,8 @@ class LinterView
     @linters = []
     grammarName = @editor.getGrammar().scopeName
     for linter in @allLinters
-      if (_.isArray(linter.syntax) and grammarName in linter.syntax or
-          _.isString(linter.syntax) and grammarName is linter.syntax or
+      if (lodash.isArray(linter.syntax) and grammarName in linter.syntax or
+          lodash.isString(linter.syntax) and grammarName is linter.syntax or
           linter.syntax instanceof RegExp and linter.syntax.test(grammarName))
         @linters.push(new linter(@editor))
 
@@ -61,7 +61,7 @@ class LinterView
         throttleInterval = parseInt(lintOnModifiedDelayMS)
         throttleInterval = 1000 if isNaN throttleInterval
         # create throttled lint command
-        @throttledLint = (_.throttle @lint, throttleInterval).bind this
+        @throttledLint = (lodash.throttle @lint, throttleInterval).bind this
 
     @subscriptions.add atom.config.observe 'linter.lintOnChange',
       (lintOnModified) => @lintOnModified = lintOnModified
--- a/lib/linter-view.coffee
+++ b/lib/linter-view.coffee
@@ -179,8 +179,7 @@ class LinterView
       lNum = message.line
       line = lines[lNum] || { 'level': -1 }
       msg_level = levels.indexOf(message.level)
-      if msg_level <= line.level
-        continue
+      continue unless msg_level > line.level
       line.level = msg_level
       line.msg = message
       lines[lNum] = line
@@ -192,12 +191,12 @@ class LinterView
 
     return unless @editor.isAlive()
 
-    if not (@showGutters or @showHighlighting)
+    unless @showGutters or @showHighlighting
       @updateViews()
       return
 
     @markers ?= []
-    for lNum, line of @sortMessagesByLine(messages)
+    for _, line of @sortMessagesByLine(messages)
       marker = @createMarker(line.msg)
       @markers.push marker
 
--- a/lib/linter-view.coffee
+++ b/lib/linter-view.coffee
@@ -89,6 +89,11 @@ class LinterView
         @showHighlighting = showHighlighting
         @display()
 
+    @subscriptions.add atom.config.observe 'linter.hushInfoMessages',
+      (hushInfoMessages) =>
+        @hushInfoMessages = hushInfoMessages
+        @display()
+
   # Internal: register handlers for editor buffer events
   handleEditorEvents: =>
     @editor.onDidChangeGrammar =>
@@ -171,10 +176,12 @@ class LinterView
     return marker
 
   # Internal: Pidgeonhole messages onto lines. Each line gets only one message,
-  # the message with the highest level presides.
+  # the message with the highest level presides. Messages of unrecognizable
+  # level (or hushed by config) will be skipped.
   sortMessagesByLine: (messages) ->
     lines = {}
     levels = ['info', 'warning', 'error']
+    levels.shift() if @hushInfoMessages
     for message in messages
       lNum = message.line
       line = lines[lNum] || { 'level': -1 }
--- a/lib/linter-view.coffee
+++ b/lib/linter-view.coffee
@@ -142,7 +142,7 @@ class LinterView
   # Internal: Process the messages returned by linters and render them.
   #
   # messages - An array of messages to annotate:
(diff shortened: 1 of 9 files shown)


park9140:
@beck, want to add a few tests for this?

beck:
@park9140 - I've added some tests.

Testing here was a bit rough.  This area of code seems to be pretty sparse in terms of testing.

I wish the dependency injection of the editor was done differently, a bit more quarantined.  Since the editor is deeply coupled, (and trying not to avoid invoking the workspace) the test setup and tear downs required some heaving stubbing to get a `LinterView` instance.

I rewrote the pull.  Breaking `display` into smaller routines with associated unit tests.

There is definitely more cleanup that can happen in `LinterView` but the section I touched now has better coverage.

spec/linter-view-spec.coffee
```diff
@@ -41,3 +46,55 @@ describe "LinterView:lint", ->
...
+    sinon.stub(LinterView.prototype, 'updateViews')
+    lv = new LinterView(stub_editor)
+
+  afterEach ->
+    LinterView.prototype.initLinters.restore()
+    LinterView.prototype.handleEditorEvents.restore()
+    LinterView.prototype.updateViews.restore()
+
+
+  describe "message sorrting", ->
```
park9140: :pencil2:  `sorting`

park9140:
@dmnd, any thoughts on this?  I've been thinking this has been missing for a while. Would be nice to get in.

dmnd:
I haven't reviewed the code, so I could be making assumptions here (sorry), but I'm not really that enthusiastic about a simple integration of info level messages if we don't have a reliable way to dismiss them. Consider what the different levels of lint message mean:

 * Error: 99% of the time I will fix this and the message will disappear
 * Warning: ~97% of the I will adhere to the message and it will disappear
 * Info: often the message should be ignored and so it will be persistently present in my editor

For error and warnings, if I don't agree with the linter, I'm usually happy with disabling that lint message entirely. Very infrequently I will agree with the class of lint message, but decide it doesn't apply to the specific example in my code. I add a comment to disable the message locally in these cases, which is ugly, but infrequent.

But info level warnings are often useful but just as often not. I don't want to disable the entire class of message because that defeats the purpose of in "info" message. But I also don't want to little my code with local disabling comments. The only other option is to put up with info messages persistently being present in my editor.

Perhaps if there is a quick way to disable/enable info messages I'll be happier with them. I do agree they are useful, but by their nature I don't think they should be as persistent as errors and warnings, and right now all the different kinds of views (inline, gutter, highlight) all assume you always want to know about a message.

lib/linter-view.coffee
```diff
@@ -142,48 +142,64 @@ class LinterView
...
-        else if message.level == 'warning'
-          'linter-warning'
-        continue unless klass?  # skip other messages
-
-        marker = @editor.markBufferRange message.range, invalidate: 'never'
-        @markers.push marker
-
-        if @showGutters
-          @editor.decorateMarker marker, type: 'gutter', class: klass
+    if not (@showGutters or @showHighlighting)
```
dmnd: nit: `unless @showGutters or @showHighlighting` is probably more idiomatic here

lib/linter-view.coffee
```diff
@@ -142,48 +142,64 @@ class LinterView
...
+      @updateViews()
+      return
 
-        if @showHighlighting
-          @editor.decorateMarker marker, type: 'highlight', class: klass
+    @markers ?= []
+    for lNum, line of @sortMessagesByLine(messages)
```
dmnd: nit: `lNum` seems to be unused so call it `_`

lib/linter-view.coffee
```diff
@@ -142,48 +142,64 @@ class LinterView
...
     return unless @markers?
     m.destroy() for m in @markers
     @markers = null
 
-  # Internal: Render all the linter messages
-  display: ->
+  # Internal: Create marker from message
+  createMarker: (message) ->
+    marker = @editor.markBufferRange message.range, invalidate: 'never'
+    klass = 'linter-' + message.level
```
dmnd: I think some linters return levels other than ['error', 'warning', 'info'], so you should probably be more restrictive here so we don't have unstyled markers

lib/linter-view.coffee
```diff
@@ -142,48 +142,64 @@ class LinterView
...
+  # the message with the highest level presides.
+  sortMessagesByLine: (messages) ->
+    lines = {}
+    levels = ['info', 'warning', 'error']
+    for message in messages
+      lNum = message.line
+      line = lines[lNum] || { 'level': -1 }
+      msg_level = levels.indexOf(message.level)
+      if msg_level <= line.level
+        continue
```
dmnd: nit: `continue if...`

could also perhaps make this `continue unless msg_level > line.level`, but that's your call

stylesheets/gutter.atom-text-editor.less
```diff
@@ -2,6 +2,15 @@
...
 .gutter .line-number {
+  &.linter-info .icon-right {
+    visibility: visible;
+    &:before {
+      color: @text-color-info;
+      content: @primitive-dot;
+      font-size: 1.1em;
+    }
+  }
+
```
dmnd: can you please post some screenshots of this?

spec/linter-view-spec.coffee
```diff
@@ -41,3 +46,55 @@ describe "LinterView:lint", ->
     waitsFor ->
       calls is linterClasses.length
     , "lint file finished"
+
+
+describe "LinterView", ->
```
dmnd: Thanks a tonne for digging in to test this stuff. Definitely a better state than you found it.

spec/linter-view-spec.coffee
```diff
@@ -41,3 +46,55 @@ describe "LinterView:lint", ->
...
+
+  describe "message sorrting", ->
+
+    it "selects the most severe message for each line", ->
+      lines = lv.sortMessagesByLine(messages)
+      expect(lines['1'].level).to.equal(2)
+      expect(lines['3'].level).to.equal(0)
+      expect(lines['9'].level).to.equal(1)
+
+    it "ignores messages of unrecognized levels", ->
```
dmnd: ...not only did you not check for this when I initially thought you missed it, you even have a test! :heart:

spec/linter-view-spec.coffee
```diff
@@ -41,3 +46,55 @@ describe "LinterView:lint", ->
...
+      lines = lv.sortMessagesByLine(messages)
+      expect(lines).to.not.property('20')
+
+
+  describe "message display", ->
+
+    it "bails if gutters and highlighting is turned off", ->
+      lv.showGutters = lv.showHighlighting = false
+      lv.display(messages)
+      expect(lv.makres).to.not.exist
```
dmnd: typo: `makres`

park9140:
@dmnd, at this point there are not any linters producing info level errors, so it is unlikely to produce an overwhelming amount of issues by adding it.  That said I do think we will needs some way of configuring what level of messaging we care about.

beck:
@dmnd how you interact with errr/warning/info is the exact same as my own behavior.

Let me explain the motivation then for introducing `info`.  I've built a linter for python docstrings but it is far more annoying than useful because the feedback is styled as warning or errors. Introducing `info` provides a distinction.

The default styling for `info` is more subdued, making it easier to ignore my docstring violations when I don't care.  Screenshot attached above.

The hope is that linters will not start throwing out tons of `info` but use the level with as much discretion as `error` and `warning`. Or, as the case with my docstring linter, set the default level to `info` so the entire linter can be disable when annoying.

This pull also fixes a subtle bug.  Before: if a line has multiple messages, the marker will be styled according to the last message received.  Now: the message with highest priority is marked.

Thanks for the review. This is actually my first time diving into CoffeeScript. Any additional nitpicks / advice on writing idiomatic code is much appreciated.

dmnd:
> at this point there are not any linters producing info level errors

Even one linter producing info messages means that you're likely to end up with a persistent message, though. 

> The default styling for info is more subdued, making it easier to ignore when I don't care

That's a good start, but I think we need something better instead.

How about this: 
  * info messages don't show up in the gutter, or as underlines, until they are toggled on via a command `linter.toggle-info`. 
  * `linter.toggle-info` is bound to keyboard shortcut `ctrl-cmd l, i`. 
  * a count of all info messages for the current file *does* appear by default in the new status bar summary
  * clicking the blue dot in the status bar summary activates `linter.toggle-info`

What do you guys think? Does that sound like a reasonable UX for info messages?

By the way, I haven't used that docstring linter yet, but since you brought it to my attention I'm definitely going to try it - looks useful! :smile:

park9140:
@dmnd, I like your suggestions. For those of us who like to burn our retinas with more information than is actually necessary,  it would be nice to have a way of forcing the info errors on at all times.

beck:
IMO adding an info display toggle command is slightly overengineering a solution to a problem that does not (yet) exist.

I've added a config to "hush" info-level messages.  "hush" meaning hide the gutter marker and highlighting but keeping the message in the status bar.

Due to purely selfish reasons, I've toggled `hushInfoMessages` to `false` - reason being that my own linter's default level is `info`. If other linters start spamming the UI with `info` messages, then it will be needed to be defaulted to `true` (I personally hope `info` will be used with discretion).

@dmnd what is this "new status bar summary" you so speak of?

The changes to config I introduced are a bit simpler than @dmnd's suggestion but it should address the main concern: the ability to hide noise.

dmnd:
> what is this "new status bar summary"

See #277. It shows a count of each type of message for the current file. You can see it in the bottom right of the screenshot on that PR.

> a problem that does not (yet) exist... I personally hope info will be used with discretion

I think this is why we disagree. I think we have different ideas about how info will be used. 

I don't expect any linter package to filter messages generated by the underlying linter. Exercising discretion is not really the job of the linter package. Maybe your linter is different (I haven't tried it yet), but most linter packages simply conveniently display in the editor whatever output a linter executable generates. The package doesn't exercise any editorial behaviour because each underlying linter has its own config system.

As an example, when this info level ships, I expect the existing pylint linter will map its "convention" messages as "info", because they're basically synonyms. Right now I think it maps both "fatal" and "error" to our "error", and "warning" to "warning". Were you thinking that linter-pylint would only report a subset of "info" messages through to the UI?

beck:
> I don't expect any linter package to filter messages generated by the underlying linter

I see what you're saying.  We're on the same page now, and I agree.

> Were you thinking that linter-pylint would only report a subset of "info" messages through to the UI?

Nope, I'd imagine it would pipe out all convention messages as `info` just as you except.

I've removed the last commit, `hushInfoMessages` is now defaulted to `true`.  I'm thinking that this is still not good enough, the message needs to also be suppressed from the status bar too.

beck:
I've updated pull.  The config now has `showInfoMessages` which is defaulted to `false`.

I still need to update the status bar.  Still sorting out the best approach.  Tried editing `processMessage` but that had negative side-effects.  So next up is editing the statusbar view, which will have to wait until tomorrow.

beck:
Updated the status bar (with new screenshots above).

beck:
:bell:

park9140:
@dmnd, this is looking good to me, any problem with going ahead and pulling it?

dmnd:
Sorry for the wait.

And thanks for understanding my point @beck! I think it's in a good spot to ship at the moment given that the default is off. We can deal with the other stuff I mentioned if it's popular.

@park9140 feel free to pull this yourself if you want. Otherwise I promise (!) I'll dedicate some time this weekend to linter and do a release with this in it.

dmnd:
Merged, thanks @beck!

beck:
No prob.  The feedback on this pull was great, thanks all.

park9140:
@beck, Would just like to say how glad I am you put together these tests.  Our coverage has been lackluster at best and every bit helps.  Thank you.

Issue #343, opened 2015-01-15
Atom.Object.defineProperty.get is deprecated.


morganestes:
atom.workspaceView is no longer available.
In most cases you will not need the view. See the Workspace docs for
alternatives: https://atom.io/docs/api/latest/Workspace.
If you do need the view, please use `atom.views.getView(atom.workspace)`,
which returns an HTMLElement.
```
Atom.Object.defineProperty.get (/opt/homebrew-cask/Caskroom/atom/latest/Atom.app/Contents/Resources/app/src/atom.js:55:11)
StatusBarSummaryView.render (/Users/morganestes/.atom/packages/linter/lib/statusbar-summary-view.coffee:11:21)
```

morganestes:
I opened this via the one-click reporting tool inside Atom and let it populate the title and initial comment.

I just upgraded to 0.174.0, which introduced the Deprecation Cop feature to view and report anything that's calling the API differently than what is coming up with the 1.0 API (see http://blog.atom.io/2015/01/15/announcing-the-atom-1-api.html for details).

I don't know how critical it is to update, but I figured I'd go ahead and report it while I could.

dmnd:
Fixed by 526bcc5ab381f3ee2d4c057dcd6571083bc31e58

PR #351, opened 2015-01-17
Remove depricated view usage

park9140:


park9140:
@dmnd, with the move into 1.0 api of atom, we need to keep an eye out for deprecations.

dmnd:
Thanks for fixing, I just published this as 0.10.1. That issue spam was crazy, check out all the dupes on #343...

Issue #387, opened 2015-01-24
Implement atom.services based linter v2 api


park9140:
V 2.0 api spec

 - [ ] linter packages provide a 'linter' service as per https://github.com/atom/atom/issues/5165
   - linter service defines the following api on a simple object
     - `lint(selectedSubstringToBeLinted, sourceFileName)` returns promise for an array of 'lintError's or just the array if done sync
     - selector for the scopes which we should lint (will allow us to lint subsections of files)
   - `lintError`
     - `range?` atom range object defining the range to which the error applies optional if only point provided
     - `point?` atom point [row, col]  of the error if a linter can't find the range this will be expanded based on the local scope at that point
     - `message`
     - `level` error level
 - [ ] break current core linter out into separate support library
    - multiple base options 
      - `executableLinter` (provides support for executing on the command line
      - `nodeExecutable` (uses BufferedNodeProcess)
    - make base linter classes support the new linter interface
    - base linter classes support creation and management of temp files rather than core linter project
 - [ ] provide example linter project demonstrating linter api
 - [ ] provide spec/documentation
 - [ ] support linter themes, ideally provide a styleguide for theme builders to verify their theme against.

park9140:
As per #411 we should consider including documenting a theme api as part of this process.

park9140:
@dmnd, @im4x, @hd-deamon,  I updated this now that the services api is released.  I'm probably going to start building this in a week or two.  I'd love your ideas before I start.

iam4x:
Wasn't active on atom for a long time, just checked the https://github.com/atom/atom/issues/5165 it looks awesome!

I let you start, and will try to find the time for a little help next week :)

hd-deman:
@dmnd what you think about this?

hd-deman:
@dmnd  what you think about this?

dmnd:
Moving to the new services API sounds pretty good to me.

> `lint(selectedSubstringToBeLinted, sourceFileName)`

In #282 we discussed linters that lint the actual file in the filesystem instead of a temporary copy. Maybe you could have two separate methods `lintFile` and `lintString` instead of combining both of these into one method.

dmnd:
Also, I wonder if @bolinfest has opinions about this

bolinfest:
I was playing around with this in https://github.com/AtomLinter/Linter/pull/432, but I'm not happy with what's there yet.

abe33:
I'm not sure if I should comment on this one or on #432. 

[...]

PR #390, opened 2015-01-26
Add JSONiq and XQuery support

wcandillon:


dmnd:
Hmm, usually the grammar package is separate from the linter implementation. All linters live in the AtomLinter org - do you mind splitting things up?

wcandillon:
@dmnd I'd like to split things up when it's possiblr to install plugin dependency automatically. Right you need to install two plugins. You would need to install 3. What do you think?


Awesome job btw!

park9140:
@wcandillon, I am hoping to roll to a new api for linters in #387,  one of the things that has been discussed about the services api is that it may start being used as plugin dependency management.  Hopefully this will happen soon :fingers crossed:

wcandillon:
@park9140 super fancy :)

iam4x:
Since it's just a link, I'm good with it. Still pending since 5feb, I merge it :)

PR #392, opened 2015-01-26
Better differentiate the C based languages in the README.

k2b6s9j:


dmnd:
That's really repetitive and the list is already quite long.

PR #401, opened 2015-02-01
Remove varname conflict with lodash

beck:


dmnd:
oh, that's why you did that. thanks for catching this.

PR #411, opened 2015-02-04
Added more highlight styles

dajoh:
Added more highlight styles and a setting for choosing one. Also fixed 'info' regexp match not being recognized.

I don't know how the SVG squiggly line looks like on OSX, but on Windows it's barely visible.

PNG squiggly line:
png squiggly

Line style:
line

Box style:
box


park9140:
@dajoh, can you post images of all of these spanning 3 lines?

park9140:
@dajoh, also, can you post an image of what the svg line looks like, I can probably fix it in windows if I can see what you see.

dajoh:
SVG squiggly:
svg

PNG squiggly:
png

Line:
line

Box:
box

park9140:
@dajoh,  The SVG looks the same on OSX.  We could easily produce another svg that has a thicker line that will correct the visibility issue.  In fact if you open the svg in inkscape and change the stroke width it should do the trick.

I screwed up earlier and asked you for one less than the important number of lines sorry about that.  As soon as you reach 4 lines the middle portion doesn't end up with lines under under them.  There is an issue with your png and line implementations as soon as it goes to 4 lines.  Could probably release without fixing it.  Additionally the box implementation starts looking wacky as well.  You can do a similar thing as I do with the SVG version in order to get the line version to work at 4 lines and above.

Also not to push to hard.  But it would be great to have a test for the style setting code.

dmnd:
What do you guys think about publishing this as a separate package per theme? The packages will essentially be a wrapper around some CSS (an example of that is https://github.com/dmnd/pulsing-cursor).

I like the idea because it means linter core won't have to maintain n different themes. It also makes it really simple for users to learn how a highlight theme is made, because it won't be surrounded by the rest of the linter code. That might empower more users to make their own tweaks instead of being constrained by whatever is shipped with linter.

dajoh:
@park9140: I see what you mean. I've fiddled around some with the CSS and it's hard to make the box one look good. Personally though, I dont think that range errors should be rendered like this, I think that they should stop when the line ends. Doing that will probably stop Atom from combining the regions.

The straight line and PNG squiggly could probably be released as is though. Error ranges usually don't extend that far, but if they do you still have the squigglies for the start & end line, which in my opinion are the most important lines. The content inside the range may be valid (such in the case of an unterminated function call, array literal, etc) and may not warrant being highlighted.

park9140:
@dmnd, I like this suggestion.  I submitted an issue the other day about the 2.0 api for linter.  This linter theme style package you suggest could be supported by formalizing the current class structure as the supported api for themes.

It also helps to limit the responsibility of the core linter package which is going to be important while defining the 2.0 api.

dajoh:
@park9140, what's the reasoning behind making the highlighting so subtle? I haven't seen any other editors have this subtile highlighting, it's barely visible! Can you upload a picture of what it looks like with the SVG stroke thicker?

dmnd:
I don't think we need any new code in linter core to support themes. Theme styles can be tweaked to override core linter styles. Then enabling a theme is just installing its package, and disabling a theme is deactivating or uninstalling it. 

Admittedly that's a bit clunky (e.g. it will be slow to try out different themes), but I think it's fine for now. It allows us to get themes out quickly and we can deal with making a setting once we've adopted that services API stuff.

park9140:
@dajoh, When I last updated the design I wan't considering the theme.  I believe the theme I was using at the time had an overly bright red color so I opted to keep it more subdued.  I think we could easily bring the default up to have a thicker line and use one of the new design colors defined in the atom style guide to make it more appropriate to each standard syntax theme.  Any major overhauls like switching it to a line or a box view should really be done as theme packages as @dmnd suggests.

dajoh:
Alright, I'll close this pull request.

dmnd:
Thanks for raising this btw @dajoh! Are you planning on publishing those themes as packages?

PR #416, opened 2015-02-05
More visible highlights

dajoh:
Previous highlights on dark background:
prev_dark

New highlights on dark background:
new_dark

Previous highlights on light background:
prev_light

New highlights on light background:
new_light

### DIFF
--- a/images/wave.svg
+++ b/images/wave.svg
@@ -76,7 +76,7 @@
      inkscape:groupmode="layer"
      id="layer1">
     <path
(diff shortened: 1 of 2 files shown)


park9140:
@dajoh, solid cleanup,  my squiggly line is really not very consistent is it :(, while we are at it we should really take the darken off of the orbs.  It brings this in line with the base style guide more.

dajoh:
How would I go about doing that?  The opacity is inherited from .line-number.

park9140:
Bleh, nvrmnd that's bull.  Need to have a talk with the atom team about that.

Issue #412, opened 2015-02-04
Persistent Status Bar Counters


IanMitchell:
Hi, I was curious if it was possible to have the Error / Warning count on the status bar persist even when the count is 0? It looks like you're removing the elements at the moment, so I don't think I can override in my personal stylesheet. Is there another way to go about making them persist, or am I out of luck?

dmnd:
I was trying to save space by hiding these since most of the time it says "0". But I understand that it's kind of weird to have it disappear completely since having it there provides an explicit reassurance that everything is working.

Would you be happier with the following:

* Error, warning and info counts disappear when zero
* When there are no errors, warnings or info messages, a grey dot with a 0 appears to show that there are no messages at all


This way we avoid having 3 dots and 3 zeroes most of the time, but we still make the state of having no lint messages explicit.

IanMitchell:
What if instead of removing the DOM, you just use CSS to hide the warning / error elements? It would act the same way as it does now, but that way users have the ability to easily override and customize the behavior in their personal stylesheet without much maintenance / upkeep for you. I haven't extensively written Atom packages though, so not sure how difficult / feasible that solution would be.

ababol:
I like @dmnd 'solution.

But if you want @IanMitchell , we can add a custom class like '.zeroError' when there is no error, this way it would be relatively easy for you to custom or to hide the element when there is no error (thank to your custom css), so everyone is happy.

Do you agree?

IanMitchell:
I like his solution too, it just sounded like more work for you guys! I figured what I suggested would be easier to implement haha.

PR #422, opened 2015-02-12
Click status bar summary icon to move to error

### LINKED ISSUE #402: Click status bar icon to jump to error
@driskell in #123:

> If clicking the error/warning counts in the status bar did the same that would be great!

ababol:
click

Fixed https://github.com/AtomLinter/Linter/issues/402

### DIFF
--- a/lib/statusbar-summary-view.coffee
+++ b/lib/statusbar-summary-view.coffee
@@ -29,21 +29,22 @@ class StatusBarSummaryView
 
 class StatusBarSummary extends View
   initialize: (messages, editor) ->
-    @on 'click', '.linter-summary-click-container', ->
+    return unless messages.length > 0
+
+    @on 'click', ->
       moveToNextMessage messages, editor
 
   @content: (messages, editor, info, warning, error) ->
     @div class: 'linter-summary inline-block', =>
-      @div class: 'linter-summary-click-container', =>
-        if error > 0
-          @div class: 'linter-error inline-block', error, =>
-            @span class: 'icon-right'
-        if warning > 0
-          @div class: 'linter-warning inline-block', warning, =>
-            @span class: 'icon-right'
-        if info > 0
-          @div class: 'linter-info inline-block', info, =>
-            @span class: 'icon-right'
+      if error > 0
+        @div class: 'linter-error inline-block', error, =>
+          @span class: 'icon-right'
+      if warning > 0
+        @div class: 'linter-warning inline-block', warning, =>
+          @span class: 'icon-right'
+      if info > 0
+        @div class: 'linter-info inline-block', info, =>
+          @span class: 'icon-right'
 
   detached: ->
     @off 'click', '.linter-summary-click-container'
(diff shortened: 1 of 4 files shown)


lib/statusbar-summary-view.coffee
```diff
@@ -27,16 +27,24 @@ class StatusBarSummaryView
     @tile = statusBar.addRightTile({item: el, priority: 100})
 
 class StatusBarSummary extends View
+  initialize: ->
+    @on 'click', '.linter-summary-container', ->
+      console.log("moveToNextMessage")
```
ababol: Hi @dmnd , I would like to use the `moveToNextMessage` function that you made. (https://github.com/AtomLinter/Linter/blob/master/lib/linter-view.coffee#L233)

Should I move it to libs/utils.js ? Or do you prefer if I make my own function?

I would like to avoid duplicate code.

dmnd: Hey @ababol :smile: 

I'm cool with moving it to utils, though with more thought perhaps we could come up with something better. I'm wondering if there should be some kind of linter model instance that the views talk to. But I don't have a concrete suggestion at the moment. 

I encourage you to experiment with how you think we should hook this up, and report back!

Also, I noticed #424. Perhaps it's orthogonal to the thing you are working on, but I wanted to make you aware of it in case it turns out to be related.

ababol: My first thought was to make a new virtual class/view and inherit the different linter view (statusbar, inline...) from it but it's not Java ^^ and if you look at the different linter view, there are not so much duplicate/shared code between them (except the moveToNextMessage function and some little things).

And just to be sure, about https://github.com/AtomLinter/Linter/issues/424, when you are saying: `Next Message command moves the cursor in every editor, not just active editor`, it means that we should be able to go through different files not just the active one, right?

dmnd: Sorry for being unclear, I updated #424 with a description.

It's fine to start off by moving that function to utils btw. The architecture can change later if someone is motivated. Having it in utils isn't that bad.

ababol:
Hi @dmnd, this is done, so as we agreed previously, I move some function inside utils.js.

lib/statusbar-summary-view.coffee
```diff
@@ -23,20 +24,28 @@ class StatusBarSummaryView
...
-      if error > 0
-        @div class: 'linter-error inline-block', error, =>
-          @span class: 'icon-right'
-      if warning > 0
-        @div class: 'linter-warning inline-block', warning, =>
-          @span class: 'icon-right'
-      if info > 0
-        @div class: 'linter-info inline-block', info, =>
-          @span class: 'icon-right'
+      @div class: 'linter-summary-click-container', =>
```
dmnd: out of curiosity, why the extra div? did attaching the event handler to `.linter-summary` not work?

dmnd:
> Are you working on #412 at the moment?

Nope I'm not

lib/statusbar-summary-view.coffee
ababol: Yep, I am not sure why I did that.

ababol:
@dmnd I just made another commit. :wink:

PR #428, opened 2015-02-18
Push top-level require() calls down into the functions that use them.

bolinfest:
This follows the first suggestion on
https://discuss.atom.io/t/how-to-speed-up-your-packages/10903.

Previous to this diff, `linter` would take over 200ms to load on my
machine, according to Timecop. Now when I run:

    atom.packages.getLoadedPackage('linter').loadTime

I consistently see `linter` loading in under 10ms (it's usually <5ms).

Activation time has not seemed to increase dramatically as a result:

    atom.packages.getLoadedPackage('linter').activateTime

Seems to hover around 80ms.

I can probably do even better in follow-up diffs.

### DIFF
--- a/lib/init.coffee
+++ b/lib/init.coffee
@@ -1,10 +1,4 @@
-_ = require 'lodash'
-{CompositeDisposable} = require 'event-kit'
-
-LinterView = require './linter-view'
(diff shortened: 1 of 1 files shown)


dmnd:
Nice, thanks for this!

PR #429, opened 2015-02-18
Change linterViews to be an ES6 Set rather than an Array.

bolinfest:
This makes it possible to remove a dependency on `lodash` in `init.coffee`.

Curiously, this was my first justified use of the
[Embedded JavaScript](http://coffeescript.org/#embedded) feature in
CoffeeScript because I needed to use the ES6 iteration idiom for `Set`,
for which there is no equivalent in CoffeeScript.

This also updates `deactivate()` to null out references so they can be garbage collected.
This is good practice if a package toggles between activated and deactivated.


dmnd:
I've never had to use that embedded javascript feature either. Interesting!

PR #430, opened 2015-02-18
Remove event-kit dependency.

bolinfest:
Both `CompositeDisposable` and `Emitter` are available on `require('atom')`.

I even made sure that `require('event-kit').Emitter === require('atom').Emitter`
was `true`.

### DIFF
--- a/lib/linter-view.coffee
+++ b/lib/linter-view.coffee
@@ -3,7 +3,7 @@ fs = require 'fs'
 temp = require 'temp'
 path = require 'path'
 rimraf = require 'rimraf'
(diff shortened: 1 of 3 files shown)


dmnd:
:feelsgood:

bolinfest:
@dmnd Thanks for merging my changes so quickly! Would you mind doing a release in the near future so we can start leveraging these performance improvements?

dmnd:
I'll do a release tonight! Sorry for the delay. Have to set expectations after merging those other PRs so quickly ;)

PR #431, opened 2015-02-18
Make Linter load faster since all clients need to load this class.

bolinfest:
Defers the loading of lodash and XRegExp until they are used.
Replaced _.memoize() with custom logic to pave the way to removing
the lodash dependency altogether in Linter.

The only other top-level `require()` calls that could be eliminated
are for `log` and `warn`.

One thing that I found a little odd is that, assuming `@executablePath`
never changes over the lifetime of the `Linter` once it is set,
`@_statCache` will only ever have one entry in it.

### DIFF
--- a/lib/linter.coffee
+++ b/lib/linter.coffee
@@ -1,8 +1,8 @@
 fs = require 'fs'
 path = require 'path'
 {CompositeDisposable, Range, Point, BufferedProcess} = require 'atom'
-_ = require 'lodash'
-{XRegExp} = require 'xregexp'
+_ = null
+XRegExp = null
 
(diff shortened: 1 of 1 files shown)


dmnd:
> One thing that I found a little odd is that, assuming @executablePath
> never changes over the lifetime of the Linter once it is set,
> @_statCache will only ever have one entry in it.

Yeah, I think that's what [this TODO is referring to](https://github.com/AtomLinter/Linter/pull/431/files#diff-8bbae72b6288a0263998cd68a1086738R65). I don't actually know if `@executablePath` is invariant though.

PR #437, opened 2015-02-19
support clearing warning on change

spalger:
Perhaps this was undesired behavior for most, but I loved the way that sublime-linter would clear out the errors as soon as you started typing.

This change brings that feature to AtomLinter behind an off-by-default setting "Clear On Change".

image


iam4x:
I'm okay with merging, since it's an option. What do you think @hd-deman @dmnd ?

dmnd:
:+1: from me

Issue #409, opened 2015-02-03
Rename stylesheets to styles.


bolinfest:
Apparently `stylesheets/` is deprecated in favor of `styles/` for Atom 1.0. You can see the instructions for creating a new package now creates a `styles/` directory instead: https://atom.io/docs/v0.177.0/creating-a-package.

PR #446, opened 2015-02-24
Missed stylesheets reference in gruntfile

ianhattendorf:
Missed the Gruntfile for issue #409. Side note, anything that can be done about ui-variables.less not being found when linting the stylesheets?


iam4x:
@ianhattendorf good catch, and for `ui-variables.less` I tried to find something, but no clue :/

PR #463, opened 2015-03-10
feat(lintChangeMethod): add options

iam4x:
WIP


iam4x:
Okey we need to fix tests on new Atom version, specs are broken. cc @hd-deman

PR #465, opened 2015-03-11
Fix travis tests

iam4x:
WIP


iam4x:
Okey I gave it another shot, I can't find a way to use sinon correctly with `allowUnsafeEval` it keeps throwing me errors.

Maybe we should change the sinon lib to another one? Any ideas? @hd-deman @park9140

park9140:
@iam4x, we should be able to just use built in jasmine spies and stubs. 
http://jasmine.github.io/edge/introduction.html#section-Spies.

PR #481, opened 2015-03-26
:memo: atom-typescript for TypeScript

basarat:


hd-deman:
Looks like something more complex than just linter-package.

basarat:
> Looks like something more complex than just linter-package.

It is much more complex, and does a lot more. Perhaps you only want to list simple linters? If so you can close this. Would also like @park9140's opinion.

hd-deman:
why you not use [linter-tslint](https://atom.io/packages/linter-tslint) as dependency instead of implement same functionality as part of you package?

basarat:
> why you not use linter-tslint as dependency instead of implement same functionality as part of you package

I agree. I haven't reimplemented it. The functionality provided is different. One is a *linter* (tslint) the other (this one) is a `compiler` / `syntax` / `semantic` checker :heart:

park9140:
@basarat, @hd-deman, we should probably think about adding a list of external "build" tools that use the linter as their error output.  There are a couple of these packages now, and most of them depend on a core package to provide the underlaying 'linter' error messages, so they can't easily be separated from their core functionality.

Issue #427, opened 2015-02-17
Wavy underline appears to highlight previous line when editor.lineHeight is set to 1.5


adambrenecki:
Hi, I'm having some display issues with AtomLinter. When `editor.lineHeight` is unset, my editor looks like this:

![image](https://cloud.githubusercontent.com/assets/473960/6221921/da495a04-b69a-11e4-8112-59acbe388b9f.png)

However, when I set it to `1.5`, it looks like this:

![image](https://cloud.githubusercontent.com/assets/473960/6221926/f1c5f246-b69a-11e4-83b5-4ac3b3e0810c.png)

even though there is no error on the preceding line.

PR #500, opened 2015-04-13
:bug: Fix wavy underline with non-standard line height #427

ryanblak:
When line height is increased from the standard line height, wavy underlines were repeated vertically causing duplicate lines.


iam4x:
Nice thanks :+1:

PR #512, opened 2015-04-25
add `linter-js-standard`

feross:
https://atom.io/packages/linter-js-standard


iam4x:
Anddddd it's doooone :+1: 

(btw I love webtorrent)

feross:
Thanks and thanks!

PR #542, opened 2015-05-21
Iterate over base options

kevinsawicki:
Previously it didn't look like `baseOptions` was actually iterated over, since `executionTimeout` was hard-coded in there.

This pull request addresses that and also adds a helper method so that `baseOptions` and `options` use the same path.


hd-deman:
:+1:

PR #304, opened 2014-12-19
only show lint messages of the current linting file

liigo:
I'm working on Rust Linter, which would like to support Rust's package manager, Cargo. Cargo was designed to build a package (or project), which is a group of many source files. While linting a file inside a package, we build the full package (use Cargo). (You can't just compile only one of these files individually.) There are maybe some errors, cause by files other than the current linting one, which need to be ignored. This PR changes the behavior to only show lint messages of the current linting file.

This is a backward compatible improvement.

PR #552, opened 2015-06-02
Filter messages to only display those related to the current file

reem:
In certain systems, for instance the package manager cargo for
the rust language, an entire package with multiple files must
be built at once and all the errors are reported in bulk.

Previously, it was impossible to filter these messages based on the
file they apply to and only show the ones relevant to the current file.
This commit provides the hooks for linters to say which file a message
belongs to (though this is optional, to maintain backwards compatibility)
so that we can filter out all messages not related to the current file.

By default, all messages are assumed to be relevant to the current file.

This PR is an updated version of #304, which has partially bitrotted/accrued
merge conflicts, and is largely a prerequisite to productively using rust
with atom.


reem:
cc @hd-deman, who is assigned to #304

PR #555, opened 2015-06-04
Added support for multiline linter messages

ptsurko:
gjslint outputs some multiline messages that current linter displays in one line.
PS. Changes to linter-gjslint will be pulled in separate pull request.


k2b6s9j:
This is really a request that should be brought up at [linter-plus](https://github.com/AtomLinter/linter-plus).

PR #581, opened 2015-06-23
Generate temporary files for legacy linters

### LINKED ISSUE #577: Changes not recognized
I just updated `linter` to 1.0.0, and I'm getting errors due to changes not being recognized. It could be that this is a result of `linter-flake8` needing an update to match `linter`.

For instance, the top line has warning `local variable list_count is assigned but never used` and the second line has the error `undefined name set_count` after I changed the variable name from `set_count` to `list_count`. Clearly, it's used, but `linter` did not update on the change or when saved (both are checked in settings).
```python
        list_count = self._get_count_list(column_list)
        line_count = ' '.join(str(x) for x in list_count) + "\n"
```

### LINKED ISSUE #578: Old settings persist in new version
If you had configured anything in a pre-1.0 version of linter (ie. written something to `config.cson`), then those keys still exist post-1.0. This wouldn't necessarily be an issue, except that Settings View displays inputs and values for those now removed settings, which can make it seem like the new version of linter supports settings that it actually no longer does.

It might be worth adding like, a one-time activation task looking for those config keys and removing them to avoid confusion.

### LINKED ISSUE #580: Linters occasionally run
since the update my linters only run 10% of the time. Even on save they usually don't run.

AsaAyers:
Fixes #577
Fixes #578
Fixes #580

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -1,3 +1,5 @@
+fs = require('fs')
+temp = require('temp')
 
 typeMap =
   info: 'Trace'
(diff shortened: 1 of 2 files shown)


lib/legacy.coffee
```diff
@@ -1,3 +1,5 @@
+fs = require('fs')
```
steelbrain: module scope uppercase please.
It you uppercase this `fs` to `FS`, it won't conflict with local variable `fs`

iam4x: Why you would name a local variable to fs? It's a bad practice not the other way.

iam4x:
Thank's @AsaAyers for fast fix!

lib/legacy.coffee
steelbrain: @iam4x there's a variable named path right below this line and I've seen a lot of function use that name.

AsaAyers: functions shouldn't use the name `path` because it conflicts with a node module. If there were a conflict here I'd rename the function's variable, not this one.

steelbrain: You make a point.

PR #586, opened 2015-06-24
Ignore lint requests until file is saved somewhere

steelbrain:
/cc @AsaAyers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Upcoming
+
+* Lint requests are now ignored until the file is saved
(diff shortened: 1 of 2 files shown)


AsaAyers:
Would you mind updating the changelog from the PR so we don't forget? Maybe we should just keep a `# upcoming` section at the top and we'll replace it with the version when it releases?

steelbrain:
Sounds great.

steelbrain:
Done.

steelbrain:
Merging based on the discussion.

PR #590, opened 2015-06-24
Allow bottom panel to be hidden

### LINKED ISSUE #589: Settings for hidding the list of errors at the bottom
Option to toggle this list :

image

Related : https://github.com/AtomLinter/Linter/issues/579 (specifically: https://github.com/AtomLinter/Linter/issues/579#issuecomment-114679665) but deserves its own issue for tracking.

steelbrain:
Just click the active tab to hide it
Click any tab to show it again
Fixes #589

ToDo
 - [x] Also add a setting for this

### DIFF
--- a/styles/linter-plus.less
+++ b/styles/linter-plus.less
@@ -15,6 +15,9 @@ linter-message {
   overflow-y: scroll;
   max-height: 150px;
   padding: 4px 8px;
+  &[hidden]{
+    display: none;
+  }
 }
 
 #linter-inline {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Upcoming
 
 * Lint requests are now ignored until the file is saved
+* Allow the user to hide bottom panel from settings or by clicking active tab
 
 # 1.0.1
 
--- a/lib/linter-plus.coffee
+++ b/lib/linter-plus.coffee
@@ -19,6 +19,8 @@ class Linter
 
     @_subscriptions.add atom.config.observe 'linter.showErrorInline', (showBubble) =>
       @views.setShowBubble(showBubble)
+    @_subscriptions.add atom.config.observe 'linter.showErrorPanel', (showPanel) =>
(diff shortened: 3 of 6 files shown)


k2b6s9j:
Looks great! :shipit:

AsaAyers:
Looks great. It works and you even persisted whether it's hidden :+1: :+1:

AsaAyers:
Changelog and then merge?

steelbrain:
Okie dokie

iam4x:
Perfect thank's @steelbrain !

styles/linter-plus.less
```diff
@@ -15,6 +15,9 @@ linter-message {
   overflow-y: scroll;
   max-height: 150px;
   padding: 4px 8px;
+  &[hidden]{
```
ssorallen: This is the native styling for the `hidden` attribute; you shouldn't need to add this.

steelbrain: native styling of `[hidden]{display:none}` is overridden when we use a more direct selector and set `display:block`.

steelbrain: `<div id="test" hidden></div>`
```css
#test{ display: block; }
```
Now that test div won't be hidden anymore despite having that hidden attribute, so we have to add override it again
```css
#test{ display: block; }
#test[hidden] { display: none  }
```

ssorallen:
Aha got it, I took that out of context. Why is `display: block` needed on the `div` elements created here?

styles/linter-plus.less
ssorallen: Aha got it, I took that out of context. Why is display: block needed on the div elements created here?

steelbrain: It was added in a PR.  https://github.com/AtomLinter/linter-plus/pull/17

ssorallen: Strange, removing that `display: block` still works fine for me. Not worth picking it apart though, it's small and it works! Thanks for clarifying.

PR #591, opened 2015-06-24
Add TravisCI config and a spec

AsaAyers:


steelbrain:
Does it actually pass those specs? You are missing a lot of properties there :sweat_smile:

AsaAyers:
Yes, you can see the specs I wrote are only concerned with that one property. We should expand it and write specs for each aspect of LegacyAdapter's behavior

AsaAyers:
IMO having one giant spec that tests every aspect of legacyAdapter is not a good idea

steelbrain:
In that case LGTM.

PR #593, opened 2015-06-24
Fixed bug where defaultLevel is not used properly

code-tree:


k2b6s9j:
:+1:

steelbrain:
@k2b6s9j is `@defaultLevel` **always** defined?

k2b6s9j:
@steelbrain By default it is defined as 'error'. See: https://github.com/AtomLinter/Linter/blob/master/lib/linter.coffee#L40

steelbrain:
LGTM.

PR #598, opened 2015-06-24
Differences warning errors

### LINKED ISSUE #595: warnings are counted as errors
Hello,
first of all thanks for this irreplaceable plug-in and for the enhancements given by the last rewriting!

I've a suggestion - please give a look at this screenshot:
schermata 2015-06-24 alle 10 14 39
As you can see in linter warnings are counted as errors (but there are no errors in the file, just warnings). Not a big issue, of course, but it could be an enhancement to distinguish them from real, critical errors.

I can imagine something like "Current File 21/0 | Project 21/0 | X  21 warnings, 0 errors" in the bottom line, or something even more compact by using colors and icons instead of text.

thanks again!
Stefano

installed versions: linter 1.0.2; linter-csslint 0.0.13;

### LINKED ISSUE #597: Keymap is gone
Looks like the very convenient keyboard shortcuts to jump from one "marked as error or warning line" to another are gone, now.

Would it be possible to have them available again, please?

iam4x:
Fixes #595 
Fixes #597

screen shot 2015-06-24 at 13 03 14

### DIFF
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -27,12 +27,14 @@ class LinterViews
 
     @_bottomErrorStatus = new BottomStatus()
     @_bottomWarningStatus = new BottomStatus()
+    @_bottomSuccessStatus = new BottomStatus()
 
     @_bottomTabFile.initialize 'Current File', => @_changeTab 'file'
     @_bottomTabProject.initialize 'Project', => @_changeTab 'project'
 
     @_bottomErrorStatus.initialize type: 'error'
     @_bottomWarningStatus.initialize type: 'warning'
+    @_bottomSuccessStatus.initialize type: 'success'
 
     @_bottomErrorStatus.addEventListener 'click', ->
       atom.commands.dispatch atom.views.getView(atom.workspace), 'linter:next-error'
@@ -103,11 +105,23 @@ class LinterViews
     @_bottomTabFile.count = counts.file.error + counts.project.warning
     @_bottomTabProject.count = counts.project.error + counts.project.warning
 
+    hasActiveEditor = typeof atom.workspace.getActiveTextEditor() isnt 'undefined'
+
     @_bottomErrorStatus.count = counts.project.error
     @_bottomWarningStatus.count = counts.project.warning
-    @updateBottomStatusClasses()
 
-    hasActiveEditor = typeof atom.workspace.getActiveTextEditor() isnt 'undefined'
+    if !counts.project.error and !counts.project.warning
+      @_bottomSuccessStatus.show()
+    else
+      @_bottomSuccessStatus.hide()
+
+    if hasActiveEditor
+      @updateBottomStatusClasses()
+    else
+      @_bottomErrorStatus.hide()
+      @_bottomWarningStatus.hide()
+      @_bottomSuccessStatus.hide()
+
     @_bottomTabFile.visibility = hasActiveEditor
     @_bottomTabProject.visibility = hasActiveEditor
 
@@ -165,6 +179,10 @@ class LinterViews
     @_statusTiles.push statusBar.addLeftTile
       item: @_bottomWarningStatus,
       priority: -998
+
+    @_statusTiles.push statusBar.addLeftTile
+      item: @_bottomSuccessStatus,
+      priority: -997
 
   # this method is called on package deactivate
   destroy: ->
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -14,9 +14,9 @@ class LinterViews
 
     @_bottomTabFile = new BottomTab()
     @_bottomTabProject = new BottomTab()
+
     @_panel = document.createElement 'div'
     @_bubble = null
-    @_bottomStatus = new BottomStatus()
 
     @_bottomTabFile.initialize("File", =>
       @_changeTab('file')
@@ -24,9 +24,21 @@ class LinterViews
     @_bottomTabProject.initialize("Project", =>
       @_changeTab('project')
     )
-    @_bottomStatus.initialize()
-    @_bottomStatus.addEventListener 'click', ->
+
+    @_bottomErrorStatus = new BottomStatus()
+    @_bottomWarningStatus = new BottomStatus()
+
+    @_bottomTabFile.initialize 'Current File', => @_changeTab 'file'
+    @_bottomTabProject.initialize 'Project', => @_changeTab 'project'
+
+    @_bottomErrorStatus.initialize type: 'error'
+    @_bottomWarningStatus.initialize type: 'warning'
+
+    @_bottomErrorStatus.addEventListener 'click', ->
       atom.commands.dispatch atom.views.getView(atom.workspace), 'linter:next-error'
+    @_bottomWarningStatus.addEventListener 'click', ->
+      atom.commands.dispatch atom.views.getView(atom.workspace), 'linter:next-warning'
+
     @_panelWorkspace = atom.workspace.addBottomPanel item: @_panel, visible: false
 
     # Set default tab to File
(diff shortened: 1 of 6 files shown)


iam4x:
@hd-deman Why in-progress? It should be ready.

steelbrain:
@iam4x Does this show all the different labels as different icons or just Warning and Error and ignores the others?

lib/linter-views.coffee
```diff
@@ -12,19 +12,24 @@ class LinterViews
...
-    @_bottomTabProject.initialize("Project", =>
-      @_changeTab('project')
-    )
-    @_bottomStatus.initialize()
-    @_bottomStatus.addEventListener 'click', ->
+
+    @_bottomErrorStatus = new BottomStatus()
+    @_bottomWarningStatus = new BottomStatus()
+
+    @_bottomTabFile.initialize 'Current File', => @_changeTab 'file'
```
steelbrain: the previous code looks cleaner.

iam4x:
```
@_bottomTabProject.initialize("Project", =>
  @_changeTab('project')
)
```

Looks cleaner than:

```
@_bottomTabFile.initialize 'Current File', => @_changeTab 'file'
```

I'm not sure, I'm using features from coffeescript.

iam4x:
I don't know all the possible types of messages, I assume there's only `error` and `warning`. I might be wrong.

Any other type will be treated as `error` with current code.

steelbrain:
How about we keep only one indicator and rename it to something other than error?

iam4x:
@steelbrain I'm not sure about that, feels great to have error/warning splitted as @furins asked

AsaAyers:
@iam4x I updated the body of this PR to include `Fixes #595`. That will cause 595 to close as soon as this gets merged

iam4x:
I added documentation about keymaps / shortcuts to Linter commands

steelbrain:
@AsaAyers how do you feel about this? I think we will be occupying a lot of ui if we keep error and warning separate.

AsaAyers:
That's a good point. I'm not sure if we even need the words. My bar is pretty full:

selection_115

Also, I have no idea why I have errors, but they aren't showing up in the count or the labels. I don't know if it's related to this issue or not

iam4x:
@AsaAyers On master you still don't have error count in current file / project?

Use case:

- No errors / warning: show only one label with `√ ` (No text?)
- No errors with warning: Don't show errors but warning
- No warning but errors: Don't show warning but errors

Do we agree?

furins:
@AsaAyers @steelbrain I agree that words may encumber too much the bottom bar. What about using just colors and numbers (I've added also other colors like information and debug info, just to show other labels)? 
example
 
@iam4x what about when you have both warnings and errors?

AsaAyers:
That's a good point. For some reason @steelbrain wants to support an unlimited number of possible types, so this has no bounds on how wide it could be. If we're going to support types everywhere else they should be here too

iam4x:
(@furins Just edited my answer with my missing use case while you were posting)

steelbrain:
We initially had that status indicator to show the user of errors in project, but then we added tabs and it became redundant, right when I was thinking about removing it, someone asked for the feature where they want to be taken to next error or problem if you wanna call it that on click. so I kept it again.

Now adding another one besides it not something I can agree upon. Sorry.

steelbrain:
Just saw that message with colors, I am definately :+1:  on that
but we should drop the commands then

furins:
@iam4x In my (ugly) example: if no warnings/errors/anything the colored areas may collapse. When the code is just "perfect" you may put a green √

AsaAyers:
> but we should drop the commands then

What commands and why?

steelbrain:
@AsaAyers the commands like `linter:next-warning`, `linter:next-info`. We should only have one command to jump to the next option.

AsaAyers:
I agree, maybe `linter:next-message`

iam4x:
I'm okay with it, it removes complexity from commands.

steelbrain:
@iam4x It's gonna be a tiny bit complex to support all the types but I think you should do it, when `LinterViews::_render` is called, it should create a set of errors create counts of all of them, then create html elements for them and replace the old ones with them.

steelbrain:
Also @iam4x do you want to keep working on it or hand it off to someone else?

iam4x:
@steelbrain We can merge this one first, since it works. And make a new one with enhancement, what do you think?

k2b6s9j:
I have a question though, if I use like 'syntax error' as a type will it be marked as an error?

AsaAyers:
What about a compromise? There is currently a bug where it says "Error" but counts everything. Can we simply change the count to be correct, and then add the other message types as an enhancement? Basically for this PR you'd just remove the new `Warning` side of this

AsaAyers:
@k2b6s9j every `type` value on the [API](https://github.com/AtomLinter/Linter/wiki/Linter-API) becomes its own category. If you use `type: 'Syntax Error'` then it becomes a new type unrelated to `Error`. Your package also has to provide the styling for that type, which gets converted to a class named `syntax-error`

AsaAyers:
This came from a tweet I can't find now, but IMO there should be two levels. Warnings are for things that should be fixed, and Errors are for things that never should have been written.

Info is just extra noise

iam4x:
If we stick to warning / issue I can just change a bit the UI to make it smaller and we are good to go :+1:

iam4x:
Made some improvements:

screen shot 2015-06-26 at 14 57 48
screen shot 2015-06-26 at 14 57 59
screen shot 2015-06-26 at 14 58 06

cc @steelbrain @AsaAyers @k2b6s9j

steelbrain:
@iam4x Looks cool. How does it look when it has no errors?

iam4x:
screen shot 2015-06-26 at 16 53 04

Nothing is shown, should we add a green dot?

iam4x:
screen shot 2015-06-26 at 17 20 01

Green dot is live :+1:

iam4x:
TODO:

* Changelog entry
* New gif on `README.md`

steelbrain:
@iam4x how about a tick instead of a green dot, as discussed in slack.

iam4x:
I've tried, it feels weird because others errors type doesn't have any icon. A dot make it smoothier.

steelbrain:
Let me know when it's ready to check out

iam4x:
@steelbrain You can test it.

k2b6s9j:
I like it. Will it show custom types too?

steelbrain:
@k2b6s9j Negative.

k2b6s9j:
Could it? In some of my linters I have a `syntax error` type which is a lighter red. It'd be nice if that were down there too.

k2b6s9j:
Then there is `linter-clang` which has a `fatal error` type which is a darker red.

iam4x:
@k2b6s9j 
https://github.com/AtomLinter/Linter/pull/598#issuecomment-114890793
https://github.com/AtomLinter/Linter/pull/598#issuecomment-114893996

iam4x:
Guys, let's get this merged. I will rebase and then we can close 2 issues.

steelbrain:
/cc @AsaAyers @k2b6s9j @lierdakil
I am neutral on this feature, I mean, It's don't use the the status icon anyway so I'll let those who use this review this PR.

lierdakil:
Screenshots look pretty nice. What message types would be supported in final version?

k2b6s9j:
I think we should move it over to the right of the bar.

steelbrain:
@lierdakil Only `Error` and `Warning`

iam4x:
@lierdakil Like @AsaAyers said we will support split errors / warning (errors = anything else than warnings)

lierdakil:
@k2b6s9j, please don't. It'd be super-awkward on ultrawide monitors.

k2b6s9j:
@devoncarew Thoughts?

iam4x:
Left/Right side isn't in the scope of the PR, it should be in a next issue. This solves the splitting of errors and warning.

k2b6s9j:
I still think that all the different types of issues should be represented, other than that looks good to me.

lierdakil:
I'm not sure about "errors" being "everything else". Personally, I use 'error', 'warning' and 'lint' message levels... lints are not errors. Far from it.

iam4x:
We can add more types support. But the issue was splitting errors / warning :)

lierdakil:
I lean to "add more (all) types", but I guess there's also an option to limit error counter to 'error' type, and 'warning' counter to 'warning' type for now? Maybe?

steelbrain:
@iam4x sorry but I am going to close this one, because the problem was fixed by renaming `Errors` to `Issues` and the other part of the problem will be fixed very soon by having a config to set the scope of this indicator. still :clap: thanks for your efforts

iam4x:
@steelbrain No worries, I had no good purpose to support all types of "issues" can't assign automatically a color to a new type.

PR #600, opened 2015-06-24
Try using writeSync and closeSync

### LINKED ISSUE #596: Errors disappear randomly when editing
I've made a gif showing the bug:
linterbug

I've also looked at the console output and this error appears randomly (not only when errors disappear) as I make changes:
screen shot 2015-06-24 at 10 32 13

The linter in use here is linter-jscs, I haven't tried other linters yet.

AsaAyers:
Fixes #596 

@iam4x Will you try this branch and see if it fixes the problem? I can't reproduce it.

My best theory is that because `fs.write` and `fs.close` are both asynchronous it sometimes closes the file before it finishes writing. I don't know if running Linux or maybe having an SSD causes me to not run into this problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Upcoming
-* Fix a bug where messages from legacy provider
+
+* Treat legacy messages as text instead of html
+* Fix a bug causing linter messages to briefly disappear and reappear
 
 # 1.0.2
 
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -54,14 +54,13 @@ module.exports = (ClassicLinter) ->
         temp.open(tmpOptions, (err, info) ->
           return reject(err) if err
 
-          fs.write(info.fd, textEditor.getText())
-          fs.close(info.fd, (err) ->
-            return reject(err) if err
-            linter.lintFile(info.path, (results) ->
-              fs.unlink(info.path)
+          fs.writeSync(info.fd, textEditor.getText())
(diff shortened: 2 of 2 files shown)


steelbrain:
@AsaAyers where are we catching the exceptions thrown here by the way?

iam4x:
@steelbrain We are not since we are getting errors from.

AsaAyers:
now I have a try/catch so the promise will reject with any errors

steelbrain:
ChangeLog?

iam4x:
I'm not a power user of CoffeeScript, but why you don't drop the parentheses?

iam4x:
Seems fine to me not happening on the same code I was editing.

(and I don't think it's linked to system specs)
screen shot 2015-06-24 at 15 27 34

AsaAyers:
Because explicit is always better than implicit

```CoffeeScript
foo bar, baz qux
```

What does the line above mean? Is there a mistake in it? There's no possible way for you to know without knowing exactly what every variable there means. It compiles to `foo(bar, baz(qux))`, but in fact I simply missed a comma and intended `foo(bar, baz, qux)`

CoffeeScript's implicit everything is a mistake. Except for the commas because it isn't ambiguous.

```CoffeeScript
foo = [
   bar
   baz
   qux
]
```

steelbrain:
LGTM after a ChangeLog entry

iam4x:
Okey thank's @AsaAyers :+1:

PR #601, opened 2015-06-24
Fix Marker leak

aki77:
**before**

[Gyazo](http://gyazo.com/28aafe30e68030fa7b2d31ff4c0a7f80)

**after**

[Gyazo](http://gyazo.com/1f635a2044ea98041fcf385967b88305)


steelbrain:
Description please? What kind-of leak is it?

AsaAyers:
The markers weren't getting cleaned up correctly.

steelbrain:
@AsaAyers I still have no idea, but would you mind testing and merging it in and adding a changelog entry?

PR #603, opened 2015-06-24
Use a temporary directory instead of a temporary file

### LINKED ISSUE #585: Filenames appear to be being disrespected again?
Running Linter v1.0.1, I seem to have run into a possible regression of the issue fixed in #192. Specifically, linter-rubocop is once again complaining about filename-related problems in *all* Ruby files. Is this a problem with linter-rubocop itself, or does it need to be fixed in the core Linter plugin?

AsaAyers:
Fixes #585

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Treat legacy messages as text instead of html
 * Fix a bug causing linter messages to briefly disappear and reappear
 * Fix memory leak (Markers weren't getting cleaned up)
+* Use the same filename as the source when generating a temporary file (Fixes #585)
 
 # 1.0.2
 
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -1,4 +1,5 @@
 fs = require('fs')
+path = require('path')
 temp = require('temp')
 
 typeMap =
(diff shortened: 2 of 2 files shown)


lib/legacy.coffee
```diff
@@ -51,15 +52,18 @@ module.exports = (ClassicLinter) ->
...
+            tmpFile = path.join(tmpDir, path.basename(filePath))
+            fs.writeFileSync(tmpFile, textEditor.getText())
 
-            linter.lintFile(info.path, (results) ->
-              fs.unlink(info.path)
+            linter.lintFile(tmpFile, (results) ->
+              # fs.rmdir only works on empty directories, so we have to delete
+              # the file first
+              fs.unlink(tmpFile)
+              fs.rmdir(tmpDir)
```
steelbrain: are these methods synchronous?

AsaAyers: good catch. They do have to execute in that order

AsaAyers: It's fixed now

lib/legacy.coffee
```diff
@@ -51,15 +52,21 @@ module.exports = (ClassicLinter) ->
...
       return new Promise((resolve, reject) ->
-        temp.open(tmpOptions, (err, info) ->
+        temp.mkdir('AtomLinter', (err, tmpDir) ->
           return reject(err) if err
 
           try
-            fs.writeSync(info.fd, textEditor.getText())
-            fs.closeSync(info.fd)
+            tmpFile = path.join(tmpDir, path.basename(filePath))
+            fs.writeFileSync(tmpFile, textEditor.getText())
```
steelbrain: how about we make it a `writeFile` and take things deeper :smiley:

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Treat legacy messages as text instead of html
 * Fix a bug causing linter messages to briefly disappear and reappear
 * Fix memory leak (Markers weren't getting cleaned up)
+* Use the same filename as the source when generating a temporary file (Fixes #585)
```
steelbrain: should we reference issues in changelog?

AsaAyers: I don't see any reason not to. The changelog seemed incomplete and kind of arbitrary without listing why this was done

steelbrain: Cool :+1:

lib/legacy.coffee
steelbrain: I remember seeing an `io.js`'s commit where it barks when we execute a sync method as it jams the event loop for that event. I think it's best practice to use async `writeFile` here.

AsaAyers: I hate Node's callback style because it makes the code really ugly and awkward to work with.

It's supposed to stop the event loop. That's what `*Sync` means.

AsaAyers: I have an idea. I'll try it out and get back to you

steelbrain: is it to have a promise wrapper around `fs`?

steelbrain:
Did you verify that it works with these promises?

AsaAyers:
yes

steelbrain:
LGTM then.

PR #609, opened 2015-06-25
Make bubble follow cursor

steelbrain:
Just like the title says

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Upcoming
 
+* Make the bubble follow cursor
+
 # 1.0.4
 
 * Fix a critical bug introduced by 1.0.3
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -72,7 +72,7 @@ class LinterViews
     try @_messages.forEach (message) =>
       return unless message.currentFile
       return unless message.range?.containsPoint? point
-      @_bubble = activeEditor.markBufferRange(message.range, {invalidate: 'never'})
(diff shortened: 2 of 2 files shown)


iam4x:
Can you please add a boolean option to toggle this option, like autocomplete+  and then I'm good to merge :)

steelbrain:
@iam4x: I think @AsaAyers was against having an option and making this the only and default behavior.
You might wanna discuss it with him.
/cc @AsaAyers

AsaAyers:
@iam4x Can you give me a reason why you want to permanently attach the bubble to the beginning of the line? And "That's the way it works now" doesn't count.

I think start of the error is better than the end, but following the cursor is the right solution. I didn't consider multi line errors when I moved it to the beginning of the error.

iam4x:
@AsaAyers Wait for the next issue: 'yo dawg I want old behaviour back'

I'm just thinking for the users, I prefer for myself to have issue following cursor :)

steelbrain:
@iam4x I agree that people are gonna want the old behavior back, I also feel happy because you also want the same behaviour.
But we can not please everyone, and people will eventually have to get used to this new behavior. The old behavior is not the best thing from the UX point of view.

AsaAyers:
I'm totally ok with 'yo dawg I want old behaviour back', but IMO it needs to have a reason. If you don't want your software to ever change, then stop using Atom. Go use Vim or Emacs, configure it the way you want and never change again.

I'm a bit surprised with how simple this change is. :+1: :+1:

Issue #608, opened 2015-06-24
Make the Bubble Transparent when Key is Held Down


k2b6s9j:
> Another thing I liked from e.g. Visual Studio is to make it (very) translucent when holding down the Ctrl key, which is also very handy in these cases.
@poelstra from #607 

This is a great idea. We could execute this with a command to enable and disable bubble transparency.

steelbrain:
What key would you propose for that?
and if we make it a command, it won't work with **holding a key down**.
The user will have to press that combination to make it disapper or appear.
If we want a key for that we will have to hard code it. :facepalm:

iam4x:
@steelbrain Can't we use `document.addEventListener('keydown', ....)` ?

steelbrain:
@iam4x Your suggestion just triggered an idea. How about we make a hollow command, and get it's keymapping programmatically and listen to `keyup/down` with it.

But before I start working on it, I want a green signal from You (@iam4x) and @AsaAyers

AsaAyers:
I'm not convinced this is needed now that the bubble follows the cursor. You can always move the cursor down to read the next line.

Also, I think the fact that we can't use normal keymaps and will have to hard code some special event is a good reason not to do this. I don't want `ALT`, if we had this I'd use `CTRL`. But Mac users rarely seem to want to use `CTRL`, so we need to support `COMMAND`.

steelbrain:
@k2b6s9j now that the bubble follows the cursor, would you say you have changed your mind?

poelstra:
Even if the bubble follows the cursor, it can be very helpful just to be able to 'peek' at the lines below, without having to move the cursor down, then up to edit again.

Maybe it's not that bad to hardcode CTRL on Windows/Linux and COMMAND on Mac?
Like follow-cursor is also hardcoded?

I would vote against ALT though, because that usually 'activates' the menu (not in Atom, apparently, but it's pretty standard behavior otherwise).

AsaAyers:
If someone can provide a clean way to implement it I'd be less opposed. It seems like there should be a way to do it.

PR #617, opened 2015-06-25
Add command set-bubble-transparent

josa42:
I add a command to set the bubble transparent. This can be be mapped to a key. On the next keyup event the bubble is set opaque again.

```
'atom-text-editor:not(.mini)':
  'ctrl': 'linter:set-bubble-transparent'
```

This was discussed here: #608

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 * Fix message render for non TextEditor panes (Fixes #610)
 * Make the bubble follow cursor
 * Show a nicer error if linter binary doesn't exist (Fixes #612)
-* Add set-bubble-transparent to set the bubble transparent until the key is released (Fixes #608)
+* Add set-bubble-transparent command to set the bubble transparent until the key is released (Fixes #608)
 
 # 1.0.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Fix message render for non TextEditor panes (Fixes #610)
 * Make the bubble follow cursor
 * Show a nicer error if linter binary doesn't exist (Fixes #612)
+* Add set-bubble-transparent to set the bubble transparent until the key is released (Fixes #608)
 
 # 1.0.4
 
--- a/lib/commands.coffee
+++ b/lib/commands.coffee
@@ -6,7 +6,7 @@ class Commands
     @_subscriptions.add atom.commands.add 'atom-workspace',
       'linter:next-error': @nextError.bind(@)
       'linter:toggle': @toggleLinter.bind(@)
-      'linter:set-bubble-transparent': @setBubleTransparent.bind(@)
+      'linter:set-bubble-transparent': @setBubleTransparent
 
     # Default values
     @_messages = null
@@ -16,7 +16,7 @@ class Commands
     return unless activeEditorLinter
     activeEditorLinter.toggleStatus()
 
-  setBubleTransparent: ->
+  setBubleTransparent: =>
     @linter.views.setBubbleTransparent()
 
   nextError: ->
--- a/lib/commands.coffee
+++ b/lib/commands.coffee
@@ -6,6 +6,7 @@ class Commands
     @_subscriptions.add atom.commands.add 'atom-workspace',
       'linter:next-error': @nextError.bind(@)
       'linter:toggle': @toggleLinter.bind(@)
+      'linter:set-bubble-transparent': @setBubleTransparent.bind(@)
 
     # Default values
     @_messages = null
@@ -14,6 +15,9 @@ class Commands
     activeEditorLinter = @linter.getActiveEditorLinter()
     return unless activeEditorLinter
     activeEditorLinter.toggleStatus()
+
+  setBubleTransparent: ->
+    @linter.getActiveEditorLinter()?.setBubbleTransparent()
 
   nextError: ->
     if not @_messages or (next = @_messages.next()).done
--- a/lib/commands.coffee
+++ b/lib/commands.coffee
@@ -17,7 +17,7 @@ class Commands
     activeEditorLinter.toggleStatus()
 
   setBubleTransparent: ->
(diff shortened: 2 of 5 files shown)


lib/commands.coffee
```diff
@@ -15,6 +16,9 @@ class Commands
     return unless activeEditorLinter
     activeEditorLinter.toggleStatus()
 
+  setBubleTransparent: ->
+    @linter.getActiveEditorLinter()?.setBubbleTransparent()
```
steelbrain: how about calling `@linter.views.setBubbleTransparent` directly

lib/editor-linter.coffee
```diff
@@ -25,6 +25,9 @@ class EditorLinter
   toggleStatus: ->
     @setStatus !@_status
 
+  setBubbleTransparent: ->
```
steelbrain: You can just remove this and access views directly

steelbrain:
Would you mind joining us in slack? 
http://atom-slack.herokuapp.com

k2b6s9j:
:+1: We should merge after the code is cleaned up.

steelbrain:
LGTM after the changes.

lib/commands.coffee
josa42: Your right. I'll do that

josa42:
I just noticed, when the editor looses focus the bubble stays transparent. So I need a way to handle this.

steelbrain:
do this in the file require area
```coffee
currentWindow = require('remote').getCurrentWindow()
```
and this in `setTransparent`
```js
currentWindow.on 'blur', @setBubbleOpaque
```
and in `setBubbleOpaque`
```js
currentWindow.off 'blue', @setBubbleOpaque
```

More info [here][1] and [here][2].

[1]:https://github.com/atom/electron/blob/master/docs/api/browser-window.md#event-blur
[2]:https://github.com/atom/electron/blob/master/docs/api/remote.md#remotegetcurrentwindow

josa42:
@steelbrain thx

josa42:
@steelbrain Actually in this case, since it is in a view you access the window directly and use `window.addEventListener` and `window.removeEventListener`.

josa42:
I think it should be working as expected now.

steelbrain:
Confirmed

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Fix message render for non TextEditor panes (Fixes #610)
 * Make the bubble follow cursor
 * Show a nicer error if linter binary doesn't exist (Fixes #612)
+* Add set-bubble-transparent to set the bubble transparent until the key is released (Fixes #608)
```
steelbrain: How about adding `command` after that command name?

lib/commands.coffee
```diff
@@ -6,6 +6,7 @@ class Commands
     @_subscriptions.add atom.commands.add 'atom-workspace',
       'linter:next-error': @nextError.bind(@)
       'linter:toggle': @toggleLinter.bind(@)
+      'linter:set-bubble-transparent': @setBubleTransparent.bind(@)
```
AsaAyers: Stand alone `@` are discouraged in CoffeeScript. Instead, you can add a fat arrow to `setBubleTransparent` below and you don't have to bind it here.

AsaAyers: For more info see https://github.com/jashkenas/coffeescript/issues/1601

steelbrain: That one was actually my mistake (see the lines above). Noted thanks.

josa42: Oh I didn't know that. Nice!

josa42:
Do you want me to squash the commites?

lib/commands.coffee
```diff
@@ -6,6 +6,7 @@ class Commands
     @_subscriptions.add atom.commands.add 'atom-workspace',
       'linter:next-error': @nextError.bind(@)
       'linter:toggle': @toggleLinter.bind(@)
+      'linter:set-bubble-transparent': @setBubleTransparent
```
steelbrain: Well, he didn't mean that.
he meant
```coffee
'linter:set-bubble-transparent': => @setBubleTransparent
```
and can you also revert the change to fat arrow below

steelbrain:
Squashing commits is not really necessary :)

lib/commands.coffee
AsaAyers: @steelbrain is wrong. you did exactly what I meant.

lib/commands.coffee
```diff
@@ -6,6 +6,7 @@ class Commands
     @_subscriptions.add atom.commands.add 'atom-workspace',
       'linter:next-error': @nextError.bind(@)
       'linter:toggle': @toggleLinter.bind(@)
+      'linter:set-bubble-transparent': => @setBubleTransparent()
```
AsaAyers: That's fine. Both are valid options but this become less clean if parameters were involved.

```CoffeeScript
    'linter:do-something-with-parameters': (param1, param2, param3) => @something(param1, param2, param3)
```

I favor the fat arrow on the target because it's always clean.

steelbrain: @AsaAyers we will not be passing arguments so this looks good to me

AsaAyers: You're correct while still missing the point. This solution is valid here, where the other solution is valid always.

"always" being, as long as you're using CoffeeScript classes, and subclasses are created through extension and not some bizarre property copying. I could make up a stupid way to break this, but I've never seen real code that would.

steelbrain: Can we discuss this in slack please?

steelbrain:
LGTM

steelbrain:
@josa42 Thanks again. :clap:

josa42:
@steelbrain My pleasure!

steelbrain:
People, I feel like we should add a default keybinding for this. 'cause not all of users are gonna want to modify the keymap.

iam4x:
@steelbrain I'm with you!

PR #619, opened 2015-06-25
Handle error case when providers don't return results

### LINKED ISSUE #618: TypeError: undefined is not a function
On linting my `init.coffee`, the Linter throws an error:

```
TypeError: undefined is not a function
  at transform (/Users/alex/.atom/packages/linter/lib/legacy.coffee:25:5)
  at /Users/alex/.atom/packages/linter/lib/legacy.coffee:75:39
```

Upon investigation, the problem is here:
```coffeescript
transform = (filePath, textEditor, results) ->
  results.map(({message, level, range}) -> # Problem here; results.map is undefined.
    [ [startLine, startCol], [ endLine, endCol ] ] = (range.serialize?() ? range)

    msg =  {
      # If the type is non-standard just pass along whatever it was
      type: typeMap[level] ? level
      text: message
      filePath: filePath
      range: [
        [ startLine, startCol],
        [ endLine, endCol]
      ]
    }

    return msg
  )
```

My `init.coffee` looks like this:
```coffeescript
if not atom?
  alert "Something's gone very wrong."
  console.log "Atom object is missing!!!"

atom.beep() # A test.
```

steelbrain:
Fixes #618

### DIFF
--- a/lib/legacy.coffee
+++ b/lib/legacy.coffee
@@ -21,6 +21,7 @@ typeMap =
   error: 'Error'
 
 transform = (filePath, textEditor, results) ->
+  return [] unless results
   results.map(({message, level, range}) ->
     [ [startLine, startCol], [ endLine, endCol ] ] = (range.serialize?() ? range)
 


k2b6s9j:
Should work well.

PR #625, opened 2015-06-26
Add linter:lint command

### LINKED ISSUE #624: Please add/restore `linter:lint` command
I'm trying to add linter API support to [ide-haskell][1] and friends, and syntax check, which doesn't make much sense as 'on fly' linter (due to relatively slow performance), can be run on both unsaved buffer or saved file directly. It could benefit from ability to run on user request, and not just on save.

I would assume this isn't the only possible use-case, but this one is close to my heart.

[1]: https://github.com/atom-haskell/ide-haskell

steelbrain:
Fixes #624


k2b6s9j:
:+1:

steelbrain:
Minor change, merging after rolling my eyes over it several times and LGTM from @k2b6s9j

lierdakil:
:+1:

PR #627, opened 2015-06-26
Art

steelbrain:
ToDo
 - [x] Support onDidChangeProjectMessages on `Linter`
 - [x] Fix a bug where an html element would be hidden from panel when shown in bubble

### DIFF
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -153,9 +153,9 @@ class LinterViews
   _renderBubble: (message) ->
     bubble = document.createElement 'div'
     bubble.id = 'linter-inline'
-    bubble.appendChild Message.fromMessage(message, false, true)
+    bubble.appendChild Message.fromMessage(message, addPath: false, cloneNode: false)
     if message.trace then message.trace.forEach (trace) ->
-      bubble.appendChild Message.fromMessage(trace, true, false)
+      bubble.appendChild Message.fromMessage(trace, addPath: true, cloneNode: false)
     bubble
 
   _renderPanel: ->
@@ -176,7 +176,7 @@ class LinterViews
         activeEditor.decorateMarker(
           marker, type: 'highlight', class: "linter-highlight #{message.class}"
         )
-      Element = Message.fromMessage(message, @_scope is 'project', true)
+      Element = Message.fromMessage(message, addPath: @_scope is 'project', cloneNode: true)
       @_panel.appendChild Element
     @updateBubble()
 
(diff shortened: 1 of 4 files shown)


lib/views/message.coffee
```diff
@@ -1,14 +1,14 @@
...
 
   attachedCallback: ->
     @appendChild Message.renderRibbon(@message)
-    @appendChild Message.renderMessage(@message)
-    @appendChild Message.renderLink(@message, @addPath) if @message.filePath
+    @appendChild Message.renderMessage(@message, @options)
+    @appendChild Message.renderLink(@message, @options) if @message.filePath
 
-  @renderLink: (message, addPath) ->
+  @renderLink: (message, {addPath}) ->
```
lierdakil: Set default `{addPath}={}`, since bad things will happen if someone calls `@renderLink(message)` (i.e. with no second parameter)

k2b6s9j:
Puuuuurrty. :shipit:

lib/views/message.coffee
```diff
@@ -31,13 +31,16 @@ class Message extends HTMLElement
     el.textContent = message.type
     el
 
-  @renderMessage: (message) ->
+  @renderMessage: (message, {cloneNode}) ->
```
lierdakil: Set default `{cloneNode}={}`, since bad things will happen if someone calls `@renderMessage(message)` (i.e. with no second parameter)

lib/views/message.coffee
steelbrain: Bad things will happen yes, but atleast they'll know that they did something they shouldn't have.

lib/linter-views.coffee
```diff
@@ -153,9 +153,9 @@ class LinterViews
   _renderBubble: (message) ->
     bubble = document.createElement 'div'
     bubble.id = 'linter-inline'
-    bubble.appendChild Message.fromMessage(message)
+    bubble.appendChild Message.fromMessage(message, addPath: false, cloneNode: false)
```
lierdakil: No need to specify `addPath`/`cloneNode` if those are false (the default), if you follow my previous advice on default :smile:

lib/views/message.coffee
steelbrain: Read my reply here https://github.com/AtomLinter/Linter/pull/627#discussion_r33384912

lib/linter-views.coffee
steelbrain: You know what, You're right.

lib/views/message.coffee
lierdakil: I would respectfully disagree that "shouldn't have" applies here. Usually with options-object, it's considered a "good practice" to have a sensible default. I would assume here that would be `false`, since at least one is `false` in 3 out of 4 instances.

steelbrain: Fixed.

lierdakil: Not exactly what I had in mind, but fine, that's a reasonable middle ground.

PR #576, opened 2015-06-23
Show StatusBar on the right

steelbrain:
Fixes #575

PR #630, opened 2015-06-26
move the status bar from the left side to the right

devoncarew:
Move the status line contribution from the left hand side to the right.

screen shot 2015-06-26 at 11 06 28 am

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Rename the status line summary (from `Errors` to `Issues`)
 * Rename the `Current File` tab to just `File`
+* Move the `No Issues` status line contribution to the right side of the status line
 
 # 1.0.6
 
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -109,12 +109,12 @@ class LinterViews
 
   # This method is called when we get the status-bar service
   attachBottom: (statusBar) ->
-    @_statusTiles.push statusBar.addRightTile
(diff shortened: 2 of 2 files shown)


k2b6s9j:
I'd rather keep the buttons on the left.

steelbrain:
@k2b6s9j Me too

k2b6s9j:
The `No Errors` is good on the right though.

steelbrain:
@k2b6s9j Yep I agree, but for some reasons AsaAyers doesn't, that's why I abandoned https://github.com/AtomLinter/Linter/pull/576

devoncarew:
I actually don't have strong opinions about which side it should be on. I do think the status line contribution should be a little more compact...

steelbrain:
We are happy to merge it in if you keep the tabs on the left and move the status icon to the right. Just like the in the old PR I referenced above.

devoncarew:
Cool, here it is:

screen shot 2015-06-26 at 11 39 49 am

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * Rename the status line summary (from `Errors` to `Issues`)
 * Rename the `Current File` tab to just `File`
+* Move the `No Issues` status line contribution to the right side of the status line
```
steelbrain: How about
```
Move the `No Issues` badge to the right side of bottom bar
```

k2b6s9j:
:shipit:

PR #635, opened 2015-06-26
Add default keymap

### LINKED ISSUE #597: Keymap is gone

steelbrain:
Fixes #597
SideNote: What does `ctrl-k` mean?


k2b6s9j:
`ctrl-k` means you press both your `Control` and `K` keys on your keyboard in one stroke. /s

keymaps/linter.cson
```diff
@@ -0,0 +1,14 @@
...
+  'ctrl-cmd-l': 'linter:lint'
+  'ctrl-cmd-e': 'linter:next-error'
+
+'.platform-linux atom-text-editor:not(.mini)':
+  'ctrl-k ctrl-l': 'linter:lint'
+  'ctrl-k n': 'linter:next-error'
+
+'.platform-windows atom-text-editor:not(.mini)':
+  'ctrl-k ctrl-l': 'linter:lint'
+  'ctrl-k n': 'linter:next-error'
```
iam4x: No end line?

iam4x:
ctrl /= cmd on mac, how to have the compatibility?

keymaps/linter.cson
steelbrain: Added. I am not used to EOLs :smiley:

steelbrain:
I just copy/pasted it from the old one, I think we should modify the keys. Matching the old ones are not really necessary.

lierdakil:
Um. Looks a little bit weird? Why is it (essentially) ctrl-alt-l and ctrl-alt-e on OSX and a sequence of keystrokes for everything else?

steelbrain:
`ctrl-n` is already used for opening a new file, do we really need a key binding for the next error, can't we just jump?

lierdakil:
ctrl-l is already used to select current/next line by core.

iam4x:
@steelbrain Look on `README.md` with #598 I added.

```
'atom-workspace':
  'linter:next-error': 'ctrl-alt-e'
  'linter:next-warning': 'ctrl-alt-w'
  'linter:toggle': 'ctrl-alt-t'
  'linter:set-bubble-transparent': 'alt'
```

lierdakil:
And I'm not convinced `linter:lint` should have a default shortcut at all, that's not a very common use-case, I don't think.

steelbrain:
So here's my solution, we only add the peek default keymap and leave the others to those who want to use them, 'cause not everyone uses those features.

lierdakil:
I suppose `next-message` something like that could be useful as well -- that's something many would use, I think. Thoughts?

steelbrain:
@lierdakil suggested key please?

lierdakil:
IDK. `alt-.`/`cmd-.`? Not sure if `cmd-.` is taken on OSX or not though.
`.` because it's "lowercase" `>`

lierdakil:
Okay, scratch that about `cmd-.`, it could be `ctrl-.` on OSX.

lierdakil:
I guess it would be better to actually ask OSX user if such shortcut makes sense or not though.

k2b6s9j:
OS X user here, what doesn't make sense?

lierdakil:
@k2b6s9j, `ctrl-.` as a shortcut for `linter:next-error`.
(`.` because on PC keyboard, shifted it's `>`)

k2b6s9j:
For me `alt-shift-,` and `alt-shift-.` make sense for previous and next error.

k2b6s9j:
`.` on my MacBook Pro shifted is still `>`

lierdakil:
Well, that makes sense on PC as well. @steelbrain ^

k2b6s9j:
It evens works on Chromebooks!

steelbrain:
Sounds good to me

k2b6s9j:
:shipit:?

PR #637, opened 2015-06-26
Add tab to show only errors in current line

### LINKED ISSUE #587: Only show Linter panel errors for selected line
I noticed that Linter recently updated to v1.0, and this is one absolutely essential feature I really miss. The pre-1.0 option to limit visible panel errors to the selected line totally rocks, as I prefer to minimize distraction within my editor. Could this be added to v1?

Thanks,
Caleb

josa42:
I added a third tab to the bottom which lets you filter the messages in the panel to show only error of the current line.

This is still work in progress. But I would like to know what you guys think of this idea.

screen-recording

Fixes #587.

### DIFF
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -135,6 +135,7 @@ class LinterViews
     if @_currentLine isnt line
       @_currentLine = line
       @_updateLineMessages()
+      @_renderPanel()
 
 
   _updateLineMessages: ->
@@ -143,12 +144,10 @@ class LinterViews
       return unless editorLinter.editor is activeEditor
 
       @_lineMessages = []
-      editorLinter.getMessages().forEach (Gen) =>
-        Gen.forEach (message) =>
-          @_lineMessages.push message if message.range?.intersectsRow @_currentLine
-
+      @_messages.forEach (message) =>
+        if message.currentFile and message.range?.intersectsRow @_currentLine
+          @_lineMessages.push message
       @_tabs.get('line').count = @_lineMessages.length
-      @_renderPanel()
 
   # This method is called when we get the status-bar service
   attachBottom: (statusBar) ->
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -160,7 +160,8 @@ class LinterViews
     @statusTiles.push statusBar.addLeftTile
       item: @tabs.get('project'),
       priority: -1000
-    @statusTiles.push statusBar.addRightTile
+    statusIconPosition = atom.config.get('linter.statusIconPosition')
+    @statusTiles.push statusBar["add#{statusIconPosition}Tile"]
       item: @bottomStatus,
       priority: 999
 
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -12,36 +12,34 @@ class LinterViews
     @_markers = []
     @_statusTiles = []
 
+    @_tabs = new Map
+    @_tabs.set 'line', new BottomTab()
+    @_tabs.set 'file', new BottomTab()
+    @_tabs.set 'project', new BottomTab()
 
-    @_bottomTabLine = new BottomTab()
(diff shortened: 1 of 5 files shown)


steelbrain:
@josa42 You might want to merge master in.

k2b6s9j:
I'd be alright merging this in if the button was a configuration option disabled by default.

k2b6s9j:
This could really be the solution for #587.

steelbrain:
:+1: For an option that's disabled by default.

steelbrain:
btw, You have accidently added `test.js` to the repo.

test.js
```diff
@@ -0,0 +1,3 @@
+
+var tes
+
```
k2b6s9j: heh?

josa42:
@steelbrain Ops I'll remove that. But like I said it's not done yet anyway.

josa42:
I am done. I think it could be merged.

steelbrain:
Looks good, gonna test it locally

steelbrain:
Be advised that I pushed a typo fix to master, so You'll have to rebase.

k2b6s9j:
:+1:

lib/linter-views.coffee
```diff
@@ -110,9 +124,32 @@ class LinterViews
       )
       throw null
 
+  updateCurrentLine: (line) ->
+    if @_currentLine isnt line
+      @_currentLine = line
+      @_updateLineMessages()
+
+
+  _updateLineMessages: ->
```
steelbrain: I think we need LintMessages for only current editor, also changing active pane triggers `LinterViews::render`

josa42: I am not sure what you mean by than.

AsaAyers:
3 tabs is way too many. As long as we're making one optional, lets just add options for each of them an you can have any combination of 0-3 tabs you'd like.

lib/linter-views.coffee
steelbrain: You are currently collecting messages from eachLinterEditor, I had suggested that we collect only active's.

steelbrain:
@AsaAyers It's off by default and is configurable by options.

AsaAyers:
I understand that, Please add 2 more options and set them on by default to show the other two tabs. So that the people who want to only show errors from the current line don't have to have 3 tabs, 2 of which they'll never use. I'm not interested in the `Project` tab, I'm going to turn that one off

josa42:
@AsaAyers I agree. Let's have an option for every tab.

lib/linter-views.coffee
josa42: That makes sense I'll look into it.

josa42:
I think it is ready to get merge now.

steelbrain:
:clap: Gonna test it now

steelbrain:
Looks great but can you make the first letter of the config values Uppercase? 'cause in settings `File` is better than `file`. Also any ideas why isn't the status icon position config not working anymore?

josa42:
@steelbrain Sorry, I overlooked the statusIconPosition option while merging. I fixed it.

k2b6s9j:
:shipit:

robneu:
Just wanted to say :heart: :heart_eyes: :100:

bjmiller:
This is terrific.  Excellent work, @josa42 and everyone!

steelbrain:
:clap: :rose:

code-tree:
Thanks heaps for this feature. Though, is there a way to set which mode is activated when Atom is opened? I'd like to have all three tabs visible, but for Atom to default to the line mode when it opens (it currently seems to default to the file mode?)

steelbrain:
@code-tree IIRC the last tab is persisted per project as part of state

code-tree:
@steelbrain Ok thanks, I just realised it is probably losing its state because of a hack in my config for a different bug fix. Sorry for the confusion.

PR #640, opened 2015-06-26
Render when Project Messages are changed

### LINKED ISSUE #639: Linter.{setProjectMessages,deleteProjectMessages} don't update linter views
Both `Linter.setProjectMessages` and `Linter.deleteProjectMessages` emit the event `did-change-project-messages`, but, since no object in the codebase is subscribed to it directly or indirectly (via `onDidChangeProjectMessages`), the UI is not updated when project messages are added through this interface. 

I can trigger a manual refresh of the linter UI via "Linter: Lint", but it's not optimal.

steelbrain:
Fixes #639

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Upcoming
 
-* Add some default keybindings
+* Add some default keybindings (Fixes #597)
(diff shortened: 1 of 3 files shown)


steelbrain:
Tiny change, self merging after confirm that it works.

PR #641, opened 2015-06-27
Handle Project Path Changes

### LINKED ISSUE #622: Removing a remote project folder does not remove the lint errors from the pane
Admittedly, remote project folders are Nuclide-specific (see http://nuclide.io/docs/remote/), but when I remove one, I expect the lints associated with the files from such a folder to be removed from the linter's Project pane.

steelbrain:
Fixes #622

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Upcoming
 
 * Add some default keybindings (Fixes #597)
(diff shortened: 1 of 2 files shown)


steelbrain:
I miss those massive changes I used to make. :grinning:

steelbrain:
This is a tiny one so gonna self merge.

PR #642, opened 2015-06-27
Make Status Icon Position Configurable

steelbrain:


steelbrain:
It's not something that people with disagree with, as it's a compromise between left and right. so merging in after confirming no bugs.

PR #643, opened 2015-06-27
Remove underscores from the CodeBase

steelbrain:
Linter Plus originally started out by copy/pasting code from my atom-hack and wrapping it in services, It's codebase was quite rough. So this is the Plan I followed to fix it.

- Prefix internal properties
- By doing the step above, we had to replace a lot of direct changes into other module's properties with observers and setter/getters
- Remove the prefixes added in first step

This is the last step of the Plan.

/cc @joefitzgerald

### DIFF
--- a/lib/commands.coffee
+++ b/lib/commands.coffee
@@ -2,8 +2,8 @@
 
 class Commands
   constructor: (@linter) ->
-    @_subscriptions = new CompositeDisposable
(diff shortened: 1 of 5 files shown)


park9140:
:ship:

joefitzgerald:
:+1:

PR #644, opened 2015-06-27
Disabling underline shouldn't disable gutter

steelbrain:
Oops


jensljungblad:
Looks good :+1:

steelbrain:
A release has been published with this fix.

PR #645, opened 2015-06-27
Multiline messages

lierdakil:
Support for multiline messages

For reasons of backwards-compatibility, multiline messages need to have `multiline: true` field set. With `multiline: true`, `text` argument will be treated with `white-space: pre`. If `html` is set, it takes precedence over both (since I'm not entirely convinced if "multiline html" even makes sense).

Multiline messages are collapsed to show only first line by default. It is possible to expand those individually with mouse click (at any given moment, only one message can be expanded this way), or all with `linter:expand-multiline-messages` command, preferably bound to a key (`ctrl` or `shift`, for example).

For styling, there are two options I could think of. Current implementation looks something like this:

Collapsed
Collapsed

Panel message expanded
Panel message expanded

Tooltip message expanded
Tooltip message expanded

All messages expanded
All messages expanded

I've also considered an option to put marker of "expandability" at the end of first line, e.g.
Tooltip message expanded (alternative)
Tooltip message expanded (alternative)

I think first version is less ambiguous, but I'd like to hear other opinions on the subject.

If you want to test this, currently this is used by `linter` branch on https://github.com/atom-haskell/haskell-ghc-mod (bear in mind one also needs `language-haskell` and `ide-haskell` installed, and `ghc-mod` executable available on `PATH`)

### DIFF
--- a/lib/views/multiline.coffee
+++ b/lib/views/multiline.coffee
@@ -1,0 +1,21 @@
+class MessageLine extends HTMLElement
+  setText: (@textContent) -> @
+
+LineElement =
+  document.registerElement('message-line', prototype: MessageLine.prototype)
+
+class Multiline extends HTMLElement
+  attachedCallback: ->
+    @tabIndex = 0
+    for line in @text.split(/\n/)
+      if line
+        @appendChild new LineElement().setText line
+
+  setText: (@text) -> @
+
+  @fromText: (text) ->
+    new MultilineElement().setText text
+
+module.exports = MultilineElement =
+  document.registerElement('linter-multiline-message', prototype: Multiline.prototype)
+module.exports.fromText = Multiline.fromText
--- a/lib/views/multiline.coffee
+++ b/lib/views/multiline.coffee
@@ -1,15 +1,11 @@
-class MessageLine extends HTMLElement
-  setText: (@textContent) -> @
-
-LineElement =
-  document.registerElement('message-line', prototype: MessageLine.prototype)
-
 class Multiline extends HTMLElement
   attachedCallback: ->
     @tabIndex = 0
     for line in @text.split(/\n/)
       if line
-        @appendChild new LineElement().setText line
+        el = document.createElement 'linter-message-line'
+        el.textContent = line
+        @appendChild el
 
   setText: (@text) -> @
 
--- a/lib/commands.coffee
+++ b/lib/commands.coffee
@@ -7,6 +7,7 @@ class Commands
       'linter:next-error': => @nextError()
       'linter:toggle': => @toggleLinter()
       'linter:set-bubble-transparent': => @setBubbleTransparent()
+      'linter:expand-multiline-messages': => @expandMultilineMessages()
       'linter:lint': => @lint()
 
     # Default values
@@ -17,6 +18,18 @@ class Commands
 
   setBubbleTransparent: ->
     @linter.views.setBubbleTransparent()
+
+  expandMultilineMessages: ->
+    for elem in document.getElementsByTagName 'linter-multiline-message'
+      elem.classList.add 'expanded'
+    document.addEventListener 'keyup', @collapseMultilineMessages
+    window.addEventListener 'blur', @collapseMultilineMessages
+
+  collapseMultilineMessages: ->
+    for elem in document.getElementsByTagName 'linter-multiline-message'
+      elem.classList.remove 'expanded'
+    document.removeEventListener 'keyup', @collapseMultilineMessages
+    window.removeEventListener 'blur', @collapseMultilineMessages
 
   lint: ->
     try
--- a/lib/views/message.coffee
+++ b/lib/views/message.coffee
@@ -1,4 +1,5 @@
 # This file is imported in views/panel
+MultilineElement = require './multiline'
 
 class Message extends HTMLElement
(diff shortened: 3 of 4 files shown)


AsaAyers:
The arrow at the beginning is definitely more obvious. I'm not sure I'd even notice whatever the symbol is there at the end, or know what to do with it.

steelbrain:
:+1:

lib/views/multiline.coffee
```diff
@@ -0,0 +1,21 @@
+class MessageLine extends HTMLElement
```
steelbrain: Why are we inventing a custom element when a simple div would work?

lierdakil:
For the reference, "whatever symbol is there at the end" is this: https://octicons.github.com/icon/ellipsis/. Ellipsis itself is a pretty common way of indicating that there is more text available, so I figured it should be somewhat self-explanatory. Apparently not.

steelbrain:
@lierdakil He was in the favor actually.

lierdakil:
I figured I should explain what it was, since there seems to be some confusion. No offence meant or anything.

lib/views/multiline.coffee
lierdakil: Because I prefer to use `new MessageLine.setText text` instead of fumbling around with `document.createElement 'div'; ...`. I could use a `div.message-line`, but that seems like more code with no apparent benefit.

steelbrain: So you are registering a whole new element just to save a line. Woaah :smiley:

steelbrain: There is a reason we dropped the `linter-panel` and `linter-inline` Custom Elements and replaced them with traditional divs. Redundant code should be avoided wherever we can.

lierdakil: Following this reasoning, all custom elements should be removed altogether. I see no reason why.

steelbrain: If you would observe the message element, then you would observe that it contains quite a lot of code, that wouldn't be good to fit anywhere else except itself.
I am not saying that You should remove `MultilineElement`. I am however saying that You should remove `LineElement`. 'cause all it does is create a div and set it's text. which could be done in a few lines like
```
el = document.createElement 'div'
e.textContent = Content
el
```

lierdakil: Hear me out: classes make sense where several semantically different elements can have this class (that's why it's a class). If a class makes sense only applied to one semantic element, it's redundant, and can be replaced with HTML5 semantic (user-defined) element.
I could `document.createElement 'message-line'` there, but that feels weird in OO code.

LineElement doesn't increase memory complexity, since most of its methods are shared from base instances, and the rest would be allocated for any element anyway. It also doesn't affect time complexity that much. Try as I might, I can't see what's wrong with my approach... :confused:

steelbrain: IMHO, We shouldn't just create extra element classes, especially when all they do is set current text. Also there is nothing wrong with setting classes, We both have been doing it for years and haven't found a problem with it, except that you have to replace `line-element` with `.line-element`.
 and IMO The simplicity is worth the extra lines. Also that's how we are doing in the base message views and it doesn't look ugly, does it?

lierdakil: Okay, would it be satisfactory if I switch to `document.createElement 'message-line'`?

steelbrain:
How about we check the length of the message instead of adding a new field to the message object?

lierdakil:
It wouldn't be backwards-compatible. Besides, it's not length that matters, but newlines in `text`.

lib/views/multiline.coffee
steelbrain: If you would do that, then you would be using it purely for styling and nothing complex like you're doing in `MultilineElement`. It makes sense for `Multiline Element`, 'cause it has a few view logic in it. </3

steelbrain:
@lierdakil Well, You are right about those new lines, but still, we didn't know we have providers that return multi-line messages.

lierdakil:
Although applying the same treatment to overly long messages is something to consider. However, that would be out of scope for this PR.

steelbrain:
Understandable.

lierdakil:
@steelbrain: I still maintain that class would be redundant, and I hope this is acceptable.

steelbrain:
This is agreeable. :smiley:

AsaAyers:
An Ellipsis isn't obvious when it's raised to the center of the text and is surrounded by an oval. If it were along the bottom of the text then it would have been obvious.

I still think the arrows are better :)

steelbrain:
I think we all agree on arrows.

AsaAyers:
LGTM :+1:

lierdakil:
I couldn't make "proper" ellipsis (either `...` or `…`) work visually in this context, so I guess arrows it is

steelbrain:
LGTM too.

lierdakil:
:+1: :heart:

steelbrain:
:clap: :rose:

steelbrain:
A changelog entry was missing, but I think I'll push it to master.

lierdakil:
Thanks. I sort of planned to add changelog entry when this gets greenlit, but you were too fast :laughing:

steelbrain:
All is well, it's in master :smiley:

PR #651, opened 2015-06-27
Added Tests for the Configuration Options.

k2b6s9j:


steelbrain:
Do you plan on adding more? or just merge this for now?

k2b6s9j:
I'm going to add more for the rest of the options. The only test that's really 'finished' is the one for the indicator.

Too bad nothing is ever really finished. Just called a finishing point.

k2b6s9j:
Also wanted people to start poking at the test.

spec/config/lint-on-fly-spec.coffee
```diff
@@ -0,0 +1,11 @@
+describe "The Lint on the Fly Configuration Option", ->
```
AsaAyers: Why have an individual file for each config and not:


```CoffeeScript
describe "config", ->
    describe "The Lint on the Fly Configuration Option", ->
        it "is `true` by default.", ->
    describe "The Inline Tooltips Configuration Option", ->
        it "is `true` by default.", ->
```

That seems to be what Jasmine is designed for.

k2b6s9j: I honestly hadn't thought about it. Would you prefer config tests in one file, because I kinda like the smaller files.

AsaAyers: I like to keep small and similar tests in the same file.  I think the only difference between these is the `describe` / `it` text and `configString`.

Even if you combine them, I'm not opposed to splitting larger more complex tests into their own files. I can see where you might want `status-icon-position-spec` to be its own file.

As I wrote this, I don't think either approach is significantly or obviously better than the other, but those are my thoughts on it.

k2b6s9j: I'll keep smaller but similar tests in one file, but larger ones in their own.

k2b6s9j:
I'm going to go ahead and merge. I'll do more work in master.

AsaAyers:
Please don't work in master. Using PRs helps keep us aware of the changes that are happening and allows a platform for discussion if we see things that can be improved. 

Unit tests are important but not critical that they get merged quickly.

k2b6s9j:
Okay. Sounds good with me.

PR #655, opened 2015-06-28
Refactor Views codebase

### LINKED ISSUE #654: Linter panel doesn't hide when it's empty and "line" tab is selected
This is a minor UI nit.

If you're using the new "line" tab to show only current-line issues in the linter-panel, and you're on a line with no issues, but there are issues in the file, the panel doesn't hide.  It's just at its minimum height while empty.  The panel should hide in this case for consistent UI behavior.

steelbrain:
Follow up to #637
Fixes #654
/cc @josa42 

Things that could've been avoided in that PR
 - [x] Tabs are rendered on every render event
 - [x] All the tabs are shown initially
 - [x] Codebase has some redundant methods
 - [x] Some functions are un-beautiful
 - [x] Panel workspace isn't collapsed if the scope is 'line' and there are no errors

Josa42: Thank You again for your PR :bow: , We greatly appreciate your contributions. It would've been ideal to have those issues addressed in your PR but it's still cool.

New Changes
 - [x] Fix a bug where Line Messages won't be updated if we're on a line where the errors are and save. (`LinterViews::render` doesn't provide row as an argument to `updateLineMessages`)
 - [x] FIx a bug where `removeBubble` would be called twice, once in `LinterViews::render` and once in `LinterViews::updatePanel`
 - [x] Split `updatePanel` into `updatePanelMarkers` and `updatePanelMessages`
 - [x] Don't render the whole panel along with markers when all we need is to render panel's messages :racehorse:

### DIFF
--- a/lib/editor-linter.coffee
+++ b/lib/editor-linter.coffee
@@ -8,10 +8,7 @@ class EditorLinter
     @inProgress = false
     @inProgressFly = false
 
-    @lineForMessages = 0 # It's used to compare if the user has changed line so we can updateLineMessages
-
     if @editor is atom.workspace.getActiveTextEditor()
-      @lineForMessages = @editor.getCursorBufferPosition()?.row
       @linter.views.updateLineMessages(true)
 
     @emitter = new Emitter
@@ -21,11 +18,9 @@ class EditorLinter
       @editor.onDidSave => @lint(false)
     )
     @subscriptions.add(
-      @editor.onDidChangeCursorPosition ({newBufferPosition}) =>
-        lineForMessages = newBufferPosition.row
-        if lineForMessages isnt @lineForMessages
+      @editor.onDidChangeCursorPosition ({oldBufferPosition, newBufferPosition}) =>
+        if newBufferPosition.row isnt oldBufferPosition.row
           @linter.views.updateLineMessages(true)
-          @lineForMessages = lineForMessages
         @linter.views.updateBubble(newBufferPosition)
     )
     @subscriptions.add(
--- a/lib/editor-linter.coffee
+++ b/lib/editor-linter.coffee
@@ -8,7 +8,7 @@ class EditorLinter
     @inProgress = false
     @inProgressFly = false
 
-    @lineForMessages = 0 # Used to tell if the user has changed line so we can updateLineMessages
+    @lineForMessages = 0 # It's used to compare if the user has changed line so we can updateLineMessages
 
     if @editor is atom.workspace.getActiveTextEditor()
       @lineForMessages = @editor.getCursorBufferPosition()?.row
--- a/lib/editor-linter.coffee
+++ b/lib/editor-linter.coffee
@@ -8,8 +8,11 @@ class EditorLinter
     @inProgress = false
     @inProgressFly = false
 
+    @lineForMessages = 0 # Used to tell if the user has changed line so we can updateLineMessages
+
     if @editor is atom.workspace.getActiveTextEditor()
-      @linter.views.updateLineMessages(@editor.getCursorBufferPosition()?.row, true)
+      @lineForMessages = @editor.getCursorBufferPosition()?.row
+      @linter.views.updateLineMessages(true)
 
     @emitter = new Emitter
     @subscriptions = new CompositeDisposable
@@ -19,7 +22,10 @@ class EditorLinter
     )
     @subscriptions.add(
       @editor.onDidChangeCursorPosition ({newBufferPosition}) =>
-        @linter.views.updateLineMessages(newBufferPosition.row, true)
+        lineForMessages = newBufferPosition.row
+        if lineForMessages isnt @lineForMessages
+          @linter.views.updateLineMessages(true)
+          @lineForMessages = lineForMessages
         @linter.views.updateBubble(newBufferPosition)
     )
     @subscriptions.add(
--- a/lib/editor-linter.coffee
+++ b/lib/editor-linter.coffee
@@ -9,7 +9,7 @@ class EditorLinter
     @inProgressFly = false
 
     if @editor is atom.workspace.getActiveTextEditor()
-      @linter.views.updateCurrentLine @editor.getCursorBufferPosition()?.row
+      @linter.views.updateLineMessages @editor.getCursorBufferPosition()?.row
 
     @emitter = new Emitter
     @subscriptions = new CompositeDisposable
@@ -19,7 +19,7 @@ class EditorLinter
     )
     @subscriptions.add(
       @editor.onDidChangeCursorPosition ({newBufferPosition}) =>
-        @linter.views.updateCurrentLine(newBufferPosition.row)
+        @linter.views.updateLineMessages(newBufferPosition.row)
         @linter.views.updateBubble(newBufferPosition)
     )
     @subscriptions.add(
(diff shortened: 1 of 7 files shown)


lib/linter-views.coffee
```diff
@@ -131,75 +127,47 @@ class LinterViews
...
-  destroy: ->
-    @messages.clear()
-    @removeMarkers()
-    @panelWorkspace.destroy()
-    @removeBubble()
-    for statusTile in @statusTiles
-      statusTile.destroy()
+      priority: -999
 
   changeTab: (Tab) ->
```
k2b6s9j: Should this variable be lower case?

steelbrain: Yep!

lib/linter-views.coffee
```diff
@@ -131,75 +134,47 @@ class LinterViews
...
-      @renderPanel()
-
-
-  updateLineMessages: ->
-    activeEditor = atom.workspace.getActiveTextEditor()
-    @linter.eachEditorLinter (editorLinter) =>
-      return unless editorLinter.editor is activeEditor
-
-      @lineMessages = []
+  updateLineMessages: (line, shouldRender = false) ->
```
k2b6s9j: I love that shouldRender is false by default.

lib/linter-views.coffee
```diff
@@ -213,16 +188,11 @@ class LinterViews
       bubble.appendChild Message.fromMessage(trace, addPath: true)
     bubble
 
-  renderPanel: ->
-    @panel.innerHTML = ''
```
pascalpp: :+1: :D

steelbrain:
Ready for some awesome criticism people.

josa42:
@steelbrain I don' agree with everything, but it's your codebase, so it works for me.

steelbrain:
@josa42 Please tell me what you don't agree with, maybe we can come to a similar ground.

lib/linter-plus.coffee
```diff
@@ -32,6 +32,15 @@ class Linter
     @subscriptions.add atom.workspace.onDidChangeActivePaneItem =>
       @commands.lint()
 
+    ignoreTabUpdates = true
```
lierdakil: Okay, this looks horrible. If you want to only fire event on change, use `atom.config.onDidChange`.

josa42:
@steelbrain I don' want to argue about code stye or whatever you consider "un-beautiful". It just doesn't feel very nice to get a PR merged and than basically everything you did being refactored. I think most of this should have taken place in #637 before merging.

But like I said, I Am OK with it. It's your codebase.

lib/linter-plus.coffee
steelbrain: Done.

steelbrain:
@josa42 I completely agree with you, You were showing great progress there :+1: but that PR was merged when I was offline, so I didn't get the time to point them out. It's our mistake, not yours, I wasn't blaming you for anything, I was just thanking you for what you did.

lib/editor-linter.coffee
```diff
@@ -8,8 +8,11 @@ class EditorLinter
     @inProgress = false
     @inProgressFly = false
 
+    @lineForMessages = 0 # It's used to compare if the user has changed line so we can updateLineMessages
```
lierdakil: `TextEditor.onDidChangeCursorPosition` has `oldBufferPosition` argument, which I think makes this redundant.

lib/linter-views.coffee
```diff
@@ -4,49 +4,44 @@ Message = require './views/message'
...
     @messages = new Set
+    @lineMessages = new Set
     @markers = []
     @statusTiles = []
 
-    @tabs = new Map
-    @tabs.set 'line', new BottomTab()
-    @tabs.set 'file', new BottomTab()
-    @tabs.set 'project', new BottomTab()
+    @tabs = {}
```
lierdakil: Why not
```coffeescript
@tabs = 
    Line: new BottomTab()
    File: new BottomTab()
    Project: new BottomTab()
```
?

lib/editor-linter.coffee
steelbrain: Right, as always :smiley:

lib/linter-views.coffee
```diff
@@ -4,49 +4,44 @@ Message = require './views/message'
...
     @panel = document.createElement 'div'
     @bubble = null
     @bottomStatus = new BottomStatus()
 
-    @tabs.get('line').initialize 'Line', => @changeTab('line')
-    @tabs.get('file').initialize 'File', => @changeTab('file')
-    @tabs.get('project').initialize 'Project', => @changeTab('project')
+    @tabs['Line'].initialize 'Line', => @changeTab('Line')
+    @tabs['File'].initialize 'File', => @changeTab('File')
+    @tabs['Project'].initialize 'Project', => @changeTab('Project')
```
lierdakil: Much repetition here. Could be replaced by `for k,v of @tabs`

lib/linter-views.coffee
steelbrain: done

lib/linter-views.coffee
steelbrain: The value of v would change in a loop and wrapping in a self-executing function would be un-nice.

lierdakil: You could work around that hacking function scoping rules... And that's what `do` keyword is for.
```coffeescript
for k,v of @tabs
    do (k, v) =>
        v.initialize k, => @changeTab(k)
```

steelbrain: Done

lib/linter-views.coffee
```diff
@@ -131,75 +132,47 @@ class LinterViews
...
-  updateLineMessages: ->
-    activeEditor = atom.workspace.getActiveTextEditor()
-    @linter.eachEditorLinter (editorLinter) =>
-      return unless editorLinter.editor is activeEditor
-
-      @lineMessages = []
+  updateLineMessages: (shouldRender = false) ->
+    return unless @tabs['Line'].visibility
+    @lineMessages.clear()
+    activeTextEditor = atom.workspace.getActiveTextEditor()
```
lierdakil: `activeTextEditor` looks redundant to me. Seems like if `currentLine` is undefined, then whole following block boils down to nothing anyway.
Should this be condensed to `atom.workspace.getActiveTextEditor()?.getCursorBufferPosition()?.row`?

BTW, what about case of multiple cursors? Should it be `getCursorBufferPositions` and a `forEach` over those instead? Not necessarily, but something to consider.

And... it was merged. Oh well. I'll keep looking though.

steelbrain: Agreed, I think I'll push these to master
/cc @AsaAyers

AsaAyers: Sorry, I thought after 50 commits this seemed like it was done.

I still think you should open another PR instead of pushing directly to master.

steelbrain: That was just one single commit that I already did. :sweat_smile: 
https://github.com/AtomLinter/Linter/commit/cfc6cd8630f24bc51aaee3176214c5ed3981f65b

lib/linter-views.coffee
steelbrain: @lierdakil Just another reason to love ES6 https://thewebsmith.co/blog/?p=7

PR #657, opened 2015-06-28
Vertically align text in tabs with rest of status bar

johanlunds:
## Before

before

## After

after


steelbrain:
The two screenshots look identical :sweat_smile:

AsaAyers:
I have to expand the images to find it, but there is a very small difference.

k2b6s9j:
It may be more noticeable on retina displays.

Issue #510, opened 2015-04-23
At Cannot find module 'space-pen' - Failed to activate the linter package


fibric:
I've upgraded my Atom from 0.192.0 to 0.194.0 and now many packages couldn't be found.

**Atom Version**: 0.194.0
**System**: Microsoft Windows 7 Enterprise 
**Thrown From**: [linter](https://github.com/AtomLinter/Linter) package, v0.12.1

### Stack Trace

Failed to activate the linter package

```
At Cannot find module 'space-pen'

Error: Cannot find module 'space-pen'
    at Module._resolveFilename (module.js:328:15)
    at Function.Module._resolveFilename (C:\Users\denny.trebbin\AppData\Local\atom\app-0.194.0\resources\app.asar\src\module-cache.js:383:52)
    at Function.Module._load (module.js:270:25)
    at Module.require (module.js:357:17)
    at require (module.js:376:17)
    at Object.<anonymous> (C:\Users\denny.trebbin\.atom\packages\linter\node_modules\atom-message-panel\lib\MessagePanelView.js:4:7)
    at Module._compile (module.js:452:26)
    at Object.loadFile [as .js] (C:\Users\denny.trebbin\AppData\Local\atom\app-0.194.0\resources\app.asar\src\babel.js:162:21)
    at Module.load (module.js:347:32)
    at Function.Module._load (module.js:302:12)
```

### Commands

```
```

### Config

```json
{
  "core": {
    "disabledPackages": [
      "symbols-view",
      "linter-flow"
    ],
    "themes": [
      "newton-dark-ui",
      "newton-dark-syntax"
    ],
    "audioBeep": false,
    "autoHideMenuBar": true
  },
  "linter": {
    "lintDebug": true,
    "showInfoMessages": true,
    "statusBar": "Show all errors"
  }
}
```

### Installed Packages

```coffee
# User
atom-ctags, v2.7.0
atom-ternjs, v0.5.23
autocomplete-bibtex, v0.5.6
autocomplete-paths, v1.0.2
autocomplete-plus, v2.12.0
git-tab-status, v1.7.0
highlight-column, v0.4.0
highlight-line, v0.10.2
highlight-selected, v0.9.2
linter, v0.12.1
linter-coffeelint, v0.2.1
linter-csslint, v0.0.11
linter-eslint, v0.5.2
linter-jsonlint, v0.1.2
linter-stylint, v0.5.0
newton-dark-syntax, v1.1.1
[...]

PR #662, opened 2015-06-29
Catch require time errors thrown from legacy providers

steelbrain:
Ref: https://github.com/AtomLinter/Linter/issues/510


iam4x:
Look's good to me :+1:

PR #664, opened 2015-06-29
Compact Message Storage

### LINKED ISSUE #663: status-bar tabs lose configuration after Atom restart
screen shot 2015-06-29 at 08 38 46

The above picture shows my current configuration.

This works as expected:

* no tabs are shown at the status-bar
* no inline tooltip
* error is shown in bottom error pannel when I move my cursor to a line with an error

After restarting Atom, the following happens:

screen shot 2015-06-29 at 08 44 10

* configuration is still exactly the same as above
* the same error is shown as if it was on the active line, however that line isn't active, and in fact that *file* isn't active.
* it *looks* like it is showing "project wide" errors, however it is not, since the filename isn't appended to the error message, which it normally is in project-wide errors.

Only way to fix this:

* enable "show line tab in error panel"
* click on "Line 0" button to enable it
* disable "show line tab in error panel"
* things now work as expected again

steelbrain:
- [x] Remove existing messages API from EditorLinter (API Breaking </3 But no package I know uses this so it's okay)
- [x] Remove all message storage from Linter
- [x] Store messages in MessageRegistry
- [x] Deprecate but still support `{get, set, delete, observe}ProjectMessages` in favor of ``{get, set, delete, observe}Messages``
- [x] Introduce a panel container
- [x] Fix a bug where some lines won't be decorated by markers
- [x] Fix #663 
- [x] Super simplify LinterViews
- [x] Split BottomPanel into a separate view
- [x] Migrate view related config observers in LinterViews
- [x] Move message classification logic to MessageRegistry
- [x] Simplify codebase

Known bugs
 - [x] Messages are stored by linter as key, file scoped messages are replaced by active one


steelbrain:
What do you think of this idea?
/cc @AsaAyers @lierdakil @k2b6s9j

iam4x:
Nice, it makes more sense imho to have `Messages` instead of `ProjectMessages`.

What method would be made public?

steelbrain:
They are both public, but `ProjectMessages` will be removed soon.

lib/linter-plus.coffee
```diff
@@ -79,21 +73,37 @@ class Linter
...
-    @emitter.emit 'did-change-project-messages', @messagesProject
-    @views.render()
+    console.warn("Linter::setProjectMessages is deprecated, use Linter::setMessages instead")
+    return @setMessages(linter, messages)
 
   deleteProjectMessages: (linter) ->
-    @messagesProject.delete(linter)
-    @emitter.emit 'did-change-project-messages', @messagesProject
-    @views.render()
+    console.warn("Linter::deleteProjectMessages is deprecated, use Linter::deleteMessages instead")
```
iam4x: Shall we use same system of Atom? (https://atom.io/packages/deprecation-cop)

steelbrain: I was gonna switch to that before merge, these warns were temp. But still thanks for the link :clap:

iam4x: Thank's for having so much time for doing this :100:

steelbrain: :bow:

lib/linter-views.coffee
```diff
@@ -34,9 +34,8 @@ class LinterViews
 
     @panel.id = 'linter-panel'
     @updateTabs()
-
-  getMessages: ->
-    @messages
+    @linter.subscriptions.add @linter.onDidChangeMessages =>
```
lierdakil: Wait, what? Why are you adding a `Disposable` to `@linter` instead of keeping it local?

steelbrain: It's still a work in progress, I am making changes one by one. and this one is planned.

lierdakil:
I'm a little bit concerned that `Linter` is becoming a "super-class". I'm generally against super-classes. Those are a horrible pain to debug and test.

So let me propose something a little bit different: have a singleton `MessageStorage` module (see [here](https://coffeescript-cookbook.github.io/chapters/design_patterns/singleton) for a brief overview), or at least single instance of it, and handle message storage, classification, etc there.

If you choose to go singleton module route, you don't need to thread it around even, just `require` it where it's used (since Node caches `require`s, it works as a single instance).

Does that sound reasonable?

iam4x:
@lierdakil I agree with you, but I don't think that's the scope of the PR. We can do it after, I like the idea of having a store for these.

lierdakil:
@iam4x, if @steelbrain is doing it now, he might as well do it right the first time around. There is a whole lot of other possible code enhancements, so it's not like we'll run out of PR material any time soon. Enforcing encapsulation and module separation at the early stage is important, because after a while, refactoring horribly entangled parts out becomes a huge undertaking, and can lead to unavoidable API breakage, which is *bad*, for obvious reasons. That said, it's not my repo (and I prefer it this way), so you do have a final say in the matter.

steelbrain:
Guys, a lot of areas of our codebase (esp. LinterViews) are un-beautiful, they could be made better looking by rewriting them in ES6,thoughts?

steelbrain:
Only one todo to kill. :muscle:

lierdakil:
Please stick with CS, especially if you still think about getting this into core.

steelbrain:
@lierdakil I reverted all the classes I wrote in JS back to coffee. I am starting to love it :sweat_smile:

lib/views/message.coffee
```diff
@@ -43,11 +40,22 @@ class Message extends HTMLElement
...
     else if message.multiline
-      el.appendChild MultilineElement.fromText message.text
+      el.appendChild Message.processMultiLine message.text
     else
       el.textContent = message.text
     el
 
+  @processMultiLine: (text)->
+    container = document.createElement 'div'
+    container.classList.add 'linter-multiline-message'
```
lierdakil: Oh come on. We're working with HTML5 here. Why classes?

steelbrain: I didn't know you dislike classes that much :smiley: Anyways, done.

lierdakil: Because in most cases, classes are redundant. I strive for sensible minimalism in everything (not that I'm particularly good at it, but I try)

AsaAyers:
CoffeeScript is fine, provided you don't care about ever getting the ability to detect unused or undefined variables. Then you're screwed

lib/linter-plus.coffee
```diff
@@ -4,29 +4,26 @@ LinterViews = require './linter-views'
...
 
-    @subscriptions.add atom.config.observe 'linter.showErrorInline', (showBubble) =>
-      @views.setShowBubble(showBubble)
-    @subscriptions.add atom.config.observe 'linter.showErrorPanel', (showPanel) =>
-      @views.setShowPanel(showPanel)
-    @subscriptions.add atom.config.observe 'linter.underlineIssues', (underlineIssues) =>
-      @views.setUnderlineIssues(underlineIssues)
+    @messages = new Messages(this)
+    @views = new LinterViews this
+    @commands = new Commands this
```
lierdakil: Use same style please. Either have parenthesis, or don't.

Also could use `@` instead of `this` (which is more idiomatic in coffee).

lib/linter-plus.coffee
```diff
@@ -79,21 +63,37 @@ class Linter
   getLinters: ->
     @linters
 
+  setMessages: (linter, messages) ->
+    @messages.set(linter, messages)
+
+  deleteMessages: (linter, messages) ->
```
lierdakil: `messages` is unused. Also, it behaves more like `clear` than `delete`, so I would advise `clearMessages`, maybe.

lib/linter-views.coffee
```diff
@@ -1,183 +1,64 @@
...
-      Project: new BottomTab()
+BottomPanel = require('./views/bottom-panel')
+BottomContainer = require('./views/bottom-container')
+BottomStatus = require('./views/bottom-status')
+Message = require('./views/message')
 
-    @panel = document.createElement 'div'
+class LinterViews
+  constructor: (@linter) ->
+    @state = @linter.state
```
lierdakil: There is a reason why I used scaffolding for states in https://github.com/AtomLinter/Linter/pull/669. If you ever decide to store something else in state, it will kind of blow up in your face (first, it won't be compatible between versions, which can lead to a whole lot of unexpected problems for users, and second, it's better to recreate it on serialization, in case something goes wrong and state you received would be corrupted in some way).

lib/messages.coffee
```diff
@@ -0,0 +1,81 @@
...
+    @messages = new Map() # Messages = Map<Linter, Array<messages>>
+    @emitter = new Emitter
+    @subscriptions = new CompositeDisposable
+    @subscriptions.add atom.workspace.onDidChangeActivePaneItem =>
+      @messages.forEach (messages) => @classifyMessages(messages)
+      @emitter.emit 'did-classify'
+
+  set: (linter, messages) ->
+    Helpers.validateMessages(messages)
+    @classifyMessages(messages)
```
lierdakil: Classification is called synchronously here, so 'did-classify' and 'did-change' are always called together. One is redundant.

lib/messages.coffee
```diff
@@ -0,0 +1,81 @@
...
+  onDidClassify: (callback) ->
+    return @emitter.on 'did-classify', callback
+
+  classifyMessages: (messages)->
+    @count = File: 0, Project: 0
+    isProject = @linter.state.scope is 'Project'
+    activeFile = atom.workspace.getActiveTextEditor()?.getPath()
+    messages.forEach (message) =>
+      @count.Project++
+      if (not message.filePath and not isProject) or message.filePath is activeFile
```
lierdakil: Should probably maintain a `Map` with `filePath -> [Message]` and build it here. That way, don't need to re-classify on file change, only when messages have changed.

lib/linter-views.coffee
steelbrain: In that case, we're fixing it in `bottom-container:61`

lib/messages.coffee
steelbrain: Well, 'did-classify' is also called on line 18 where 'did-update' is not called.

lib/messages.coffee
steelbrain: Code example please, Your statement is a bit confusing

steelbrain:
So I removed the `editorMessages` from the storage to keep the messages from file scoped linters not available in the project tab, but then I ran into other problems like, if the editor that set it's messages as value for that linter, if it's destroyed what are we gonna do then? or if it's switched from active to non-active, that sort of complications started to arise, and what if the editor-linter is destroyed, are we gonna remove all of the file-scoped linter's messages from the message registry.
So I have decided to keep a private `@messages` var in editor-linter, but the API of `MessageRegistry` remains the same.

steelbrain:
I think I am gonna play wit this more, using the eye of performance.

lib/messages.coffee
lierdakil: Sorry about the silence.

```coffeescript
@editorMessageMap = new Map
...
@editorMessageMap.clear()
messages.forEach (message) =>
    @editorMessageMap.set message.filePath [] unless @editorMessageMap.has messageFilePath
    (@editorMessageMap.get message.filePath).push message
...
getMessagesForEditor: (editor) ->
    msgs = @editorMessageMap.get editor.buffer.getURI()
    return (msgs ? [])
```
from the top of my head, expect errors.

steelbrain:
Okay guys so I did some profiling with 266 messages and here's the results
screenshot from 2015-07-02 22 27 18

so the thing that's taking most of our time is the regex parser we use, we might wanna switch to pure js regex (/cc @AsaAyers (it's coffeelint)). other than that, It's just appendChild and render Custom Element components (Message::attachedCallback) which is a DOM thing that we can't optimize.
Please note that the machine I did this profiling on is running an ordinary HDD and 4th gen 1.9 i5 intel processor. so I think this PR is quite efficient.

steelbrain:
I need some testers people, our automated test coverage isn't very good. and I am gonna target it in my next PR where we make the modules work in isolated env.

steelbrain:
(wrong button click)

AsaAyers:
@steelbrain `linter-coffeelint` doesn't use a XRegExp, so your conclusion there is wrong. That `parse` is actually inside `coffeelint`. I think that's `JSON.parse` becuase I don't see a `parse()` in CoffeeLint.

AsaAyers:
If you really want to profile the right thing, you should use a CLI linter so the internals of how the linter runs doesn't show up in your profile

lib/helpers.coffee
```diff
@@ -1,7 +1,7 @@
 {Range} = require('atom')
 
 Helpers = module.exports =
-  validateResults: (results) ->
+  validateMessages: (results) ->
```
k2b6s9j: Here's the reason the tests are broken!

steelbrain: Fixed

steelbrain:
Merging in to keep the pace of progress (been almost a week since this is open), there have been zero known non-fixed bugs in this PR.
Gonna work on making the modules work isolated in the next PR (It'll help us write tests for them).

Issue #659, opened 2015-06-28
Issues with current implementation of status bar tabs (line, file, project)


lierdakil:
I've looked long and hard at the code, and here is a conclusion I came to:

We need to put all "tab buttons" into a container of their own, and actually create/remove individual tabs when those are enabled/disabled in config.

Why I think this would be a good idea:
1. Reduce probability of status bar conflict (when a package could insert an item between "tab buttons")
2. Ability to drop `.first` and `.last` classes in favor of `:first-of-type` and `:last-of-type` css selectors
3. Reduced memory complexity due to the fact that unnecessary tabs are not created.

What container could do besides, ahem, containing:
1. Observe config changes
2. Track changes of active tab

Thoughts?

AsaAyers:
I was actually thinking about this earlier today. One of the things I was considering was pulling all of the data storage into a central location. There is a `did-change-project-messages` that external projects can consume, but it isn't used internally and there is no file based or active tab based event. If all storage is pulled to a central location and fires events, then we only need to initialize the tabs that we use and have them subscribe to the bit of storage that matters.

It can also lead to some cleaner implementations of how messages are gathered. It might look something like this:

    projectMessages = storage.getAllMessages()
    fileMessages = projectMessages.filter(isCurrentFile)
    lineMessages = fileMessages.filter(isCurrentLine)

I think by moving storage to a central location and using events it can also help with redundant renders. I found this earlier today but haven't had time to fix it:

```CoffeeScript
  deleteLinter: (linter) ->
    return unless @hasLinter(linter)
    @linters.delete(linter)
    if linter.scope is 'project'
      # If we take this path deleteProjectMessages renders making @views.render()
     # below redundant
      @deleteProjectMessages(linter)
    else
      @eachEditorLinter((editorLinter) ->
        # this will render when it gets to the active editor, so you can't predict
        # whether it's a redundant render or not
        editorLinter.deleteMessages(linter)
      )
    @views.render()
```

lierdakil:
@AsaAyers, yes, storing messages in central location and using event subscription to notify of updates seems like a good idea. Sounds like an application for a singleton pattern.

steelbrain:
I think I'm gonna take this one

steelbrain:
Fixed by https://github.com/AtomLinter/Linter/pull/664

PR #669, opened 2015-06-29
Tabs container

### LINKED ISSUE #663: status-bar tabs lose configuration after Atom restart

lierdakil:
Not necessarily a pull request, but I'd like some comments on this. This is related to my proposition in #659 (and I probably accidentally fixed #663 along the way, which is a thing btw). You might call it a "proof-of-concept implementation". Or you might consider it a WIP.

In any case, please hold off on this until #664 is merged, I'll need to rebase on it.

### DIFF
--- a/lib/linter-plus.coffee
+++ b/lib/linter-plus.coffee
@@ -6,12 +6,10 @@ Helpers = require './helpers'
 Commands = require './commands'
 
 class Linter
-  constructor:(@state)  ->
-    @state ?= {}
-
+  constructor:(state = {})  ->
     # Public Stuff
     @lintOnFly = true # A default art value, to be immediately replaced by the observe config below
-    @views = new LinterViews @state, this # Used by editor-linter to trigger views.render
+    @views = new LinterViews state.views, this # Used by editor-linter to trigger views.render
     @commands = new Commands this
 
     # Private Stuff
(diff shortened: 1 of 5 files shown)


steelbrain:
:clap:

lib/linter-plus.coffee
```diff
@@ -34,12 +32,10 @@ class Linter
     @subscriptions.add atom.workspace.onDidChangeActivePaneItem =>
       @commands.lint()
 
-    @subscriptions.add atom.config.onDidChange 'linter.showErrorTabLine', =>
-      @views.updateTabs()
-    @subscriptions.add atom.config.onDidChange 'linter.showErrorTabFile', =>
-      @views.updateTabs()
-    @subscriptions.add atom.config.onDidChange 'linter.showErrorTabProject', =>
-      @views.updateTabs()
+    for tabName of @views.tabs.tabPriorities
```
lierdakil: Note to self: this makes little sense here, should probably move to `TabContainer`

steelbrain:
Closing in favor of #664 as discussed with @lierdakil

Issue #648, opened 2015-06-27
keybinding 'alt' collides with keybinding of atom-core


colin-kiegel:
linter adds this keybinding: **alt**	linter:set-bubble-transparent	Linter	atom-text-editor:not(.mini)

It collides with
**Atom > Settings > Core Settings > Auto Hide Menu Bar**
Automatically hide the menu bar and toggle it by pressing **Alt**. This is only supported on Windows & Linux.

This combination results in not being able to toggle the menu bar within editor windows. This might affect many users of atom, because it is a core keybinding. I suggest to change the default binding for linter to something else.

steelbrain:
What alternative key would you suggest for a peek below the bubble command?

colin-kiegel:
I don't have a preference, because I am quite new to atom (~2 hrs) and just setting it up and getting to know it :-)

k2b6s9j:
On Windows? Or everywhere?

steelbrain:
Everywhere.

k2b6s9j:
Ctrl is free on my OS X install.

k2b6s9j:
Alt was taken up by Nuclide's click to symbol.

steelbrain:
Yeah, same story on linux and windows but not sure, Can someone else confirm that?

colin-kiegel:
I tried to map CTRL on Linux, but that did not work for me for some reason (Ubuntu 15.04)

colin-kiegel:
What about switching to *toggle*-behaviour for transparency? It could then be a more complex keybinding, like CTRL+ALT+[a-Z0-9]

In this case please consider, that keyboard layouts can be quite different around the globe (mostly for special characters) - standard letters should generally be better than special characters. In fact I had to override the other keybinding alt-shift-, because shift-, is interpreted as ; with my current layout and the keybinding did not work, too. But that did not seem as relevant as this one, so I did not file it as a bug. My point is: If you are looking for new keybindings, please try to avoid shift + special-char, because that could lead to unexpected behaviour on different keyboard-layouts.

colin-kiegel:
Another idea:
* default to 'transparent bubbles' when the cursor is moved
* on 'next-error' switch to 'opaque bubbles'
** if the bubble was transparent before (i.e. the cursor had moved) than *stay* at the current position
** else go to the next error

I think that might work, too.

steelbrain:
@colin-kiegel I have just tested mapping ctrl and it works flawlessly.
Here's my keymap.cson file
[...]

PR #671, opened 2015-06-30
:new: Use alt-> to jump to next error

### LINKED ISSUE #667: Request for new keybinding instead of 'alt-shift-.': 'linter:next-error'
I suggest
```
'atom-text-editor:not(.mini)':
  'alt-ctrl-x': 'linter:next-error'
```
reasoning:
* shift + 'special char' like '.' does not work for me (linux / german keyboard layout)
* alt-ctrl-x can be used with one hand

despairblue:
Also `alt-<` is not taken and easy to reach on US and german layouts. It could be used for going to the previous error, but I think that functionality was removed and has to be reimplemented.

Fixes #667

### DIFF
--- a/keymaps/linter.cson
+++ b/keymaps/linter.cson
@@ -1,3 +1,3 @@
 'atom-text-editor:not(.mini)':
   'alt': 'linter:set-bubble-transparent'
-  'alt-shift-.': 'linter:next-error'
+  'alt->': 'linter:next-error'


AsaAyers:
@steelbrain did `previous-error` just not get implemented? If we have both they should be consistent. Otherwise, this looks fine to me

iam4x:
`previous-error` is just not implemented you are right, we use a `new Set` which has the `.next` method. I don't know if there's a `.previous` method?

despairblue:
Iterators don't have a `previous` method. As far as I understand the code from *commands.coffee* it's impossible to go back to the last error.

A Set does not seem to be the best data structure for doing that, or at least it's a bit to simple for that.

Is there the possibility that a message might be added more than once? Since the Set is cleared [here](https://github.com/AtomLinter/Linter/blob/master/lib/linter-views.coffee#L79) every render circle anyways I don't see why it would be necessary to use a Set in the first place unless the linters are allowed to add the same message more than once (with *the same* I really mean *the same object*) and can expect that only one reference is kept.

despairblue:
:bulb: I think I have an idea, maybe this works out... :pray:

despairblue:
I addressed it in #677

steelbrain:
@despairblue would you like to add a keymap entry for the previous message as well?

steelbrain:
@despairblue also would you like to address #648 in this PR?

despairblue:
I sure could, but I can't think of a good alternative of the top of my head.
Except for `ctrl` of course (which btw didn't work for me as well, **until** I reloaded atom).

@colin-kiegel have you tried reloading atom after adding this keybinding?

despairblue:
Another thing to keep in mind is that it's kind of sad to use a different modifier for the `set-bubble-transparent` and `next-error/previous-error`.

colin-kiegel:
I wonder how other IDEs handle this. I already suggested different approaches to 'opaque unless key X is held down':
(a) Toggle Opaque/Transparent and stay in that state until next toggle -> Advantage is it can be more complex than one key, because you dont need to hold it down all the time and you can type while the bubble is transparent.
(b) Opaque on explicit jump (to next error) and transparent on each manual cursor move. -> Same advantages as above.

-> I have the feeling this issue can be solved on a different level.

despairblue:
B sounds nice, but it's a bit more than only changing a key mapping.

steelbrain:
B definately sounds nice, but I have some questions:

- Will the cursor be shown when the user stops moving?
- What will be it's delay?
- Will there be an option to disable that behaviour?
- How will it work when we scroll down line?

colin-kiegel:
> Will the cursor be shown when the user stops moving?
I would suggest not, because that adds distraction. Instead consider using the next-lint shortcut: We agreed it should make bubbles opaque. What if the 'next opaque lint' equals 'current transparent lint' in this case? I think people will learn that very quickly.
> What will be it's delay?
As above, let the user decide.
> Will there be an option to disable that behaviour?
I think there should, in case you really want it in the first place.
> How will it work when we scroll down line?
I don't quite understand.

steelbrain:
Ping @despairblue 
This has been opened for a week, let's get this done

despairblue:
I won't find the time to implement anything that goes further than changing a key mapping until the end of august. I'm just too busy until then. :disappointed:

steelbrain:
All fine :)

PR #675, opened 2015-06-30
Changed priority of the bottom panel, fixes #668

### LINKED ISSUE #668: A potential UI improvement
**TL;DR**: may be possible to place the warnings/errors list above the find&replace area (or above any other UI element sticked to the bottom bar)? It seems more natural to me to have the components that may grow/shrink/collapse more frequently on the top, and the fixed-height ones to the bottom.

<hr>

This is a really small and low-priority enhancement, I know, but I'm asking just in case it could be trivial to fix. I'm trying to explain what happens with this screen capture:

find_order

Please observe the pointer: I have to move it up and down every time I'd like to replace an occurrence of text. 

In the given example, when I'm replacing quotes using the find&replace dialog the error counter grows and shrinks because of unterminated strings and this displaces also the "replace" button. I've constantly to look at "where the button is".

there are many workarounds, here are just some of them...

1. to press enter when in the "replace" text input (=use keyboards shortcuts)
2. to disable "lint on fly"
3. to toggle the lint error panel off

...but I was wondering if the linter messages may be placed above other elements of the bottom bar to avoid this "jumping effect". The reason is that, in my opinion, if an element changes its height frequently, it should be closer to the "free side" of the bar -> the upper part of a bottom bar, the left side on a right sidebar... 

This is not an issue, definitely, and may be more an atom-related enhancement than a linter one. I'm just throwing my two cents :smiley: to know your opinion about that. 

thanks in advance
Stefano

furins:
OK, I'm trying to do my homework and to contribute more actively to this plug-in, not only asking but also giving back something! :smile: 

It seems that the problem has been addressed also by Atom guys: 
https://github.com/atom/atom/issues/4834

they added [a note in workspace.coffee](https://github.com/atom/atom/blob/f73d3df2339bd2ec6017e156dbc5edadf2b9c18b/src/workspace.coffee#L670) that says:

> If your panel changes its size throughout its lifetime, consider giving it a higher
  priority, allowing fixed size panels to be closer to the edge. This allows control targets to
  remain more static for easier targeting by users that employ mice or trackpads.

Attached you'll find a PR with the change.

### DIFF
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -26,7 +26,7 @@ class LinterViews
     @bottomStatus.initialize()
     @bottomStatus.addEventListener 'click', ->
       atom.commands.dispatch atom.views.getView(atom.workspace), 'linter:next-error'
-    @panelWorkspace = atom.workspace.addBottomPanel item: @panel, visible: false
+    @panelWorkspace = atom.workspace.addBottomPanel item: @panel, visible: false, priority: 500 # priority increased because of https://github.com/AtomLinter/Linter/issues/668 
 
     for key, tab of @tabs
       do (key, tab) =>


steelbrain:
You might wanna wait for https://github.com/AtomLinter/Linter/pull/664 to land

steelbrain:
On a side note, how will it affect packages like `atom-typescript` and `atom-omnisharp` who are using their panels alongside linter's?

steelbrain:
@basarat any ideas how will it affect your panel?

basarat:
Applied the change locally and doesn't affect us: 

image

steelbrain:
Great, thanks :clap:

basarat:
We have priority `1000` don't remember why I chose that :) https://github.com/TypeStrong/atom-typescript/blob/c94788f7edb8f23fa82a606640ef297309be9437/lib/main/atom/views/mainPanelView.ts#L446

basarat:
Adding @david-driscoll just so they don't get surprised :)

radicaled:
We're good over here, too.

PR #676, opened 2015-06-30
Log Deprecations when Legacy API is used.

k2b6s9j:


k2b6s9j:
I'm not sure which location for the deprecation works better. But they both seems to do it.

k2b6s9j:
I tried to deprecate the public API methods themselves, but that placed the deprecations on Linter and the actual linter provider.

lib/linter.coffee
```diff
@@ -2,6 +2,7 @@ fs = require 'fs'
 path = require 'path'
 {CompositeDisposable, Range, Point, BufferedProcess} = require 'atom'
 XRegExp = null
+Grim = require('grim')
```
steelbrain: `{deprecate} = require('grim')`

package.json
```diff
@@ -32,6 +32,7 @@
     }
   },
   "dependencies": {
+    "grim": "^1.4.1",
```
steelbrain: You don't need to add it as a dep, It's available by default by atom I think.

lib/linter.coffee
k2b6s9j: How come?

package.json
k2b6s9j: Yeah... I thought the same, but it threw errors without the dep.

lib/linter.coffee
steelbrain: I meant that you can use destructuring, we don't need that Grim variable for any other purposes than deprecate. Then how about you extract that variable from it on require time

k2b6s9j: Ah. I see. I'll do that.

k2b6s9j:
Should be good to merge.

k2b6s9j:
And get ready for the issues to come streaming in.

steelbrain:
LGTM. Thoughts @AsaAyers

k2b6s9j:
I guess we should cc @AtomLinter/linter

AsaAyers:
I don't think this should be done until we have a complete (and verified) set of replacement tools for the CLI linters.

AsaAyers:
If I were maintaining a CLI linter, I'd say we aren't ready to migrate

k2b6s9j:
I'm maintaining a couple CLI Linters and they're working fine on the new API.

steelbrain:
Yep, if only one of us had push access of `atom-linter` npm modle

steelbrain:
@k2b6s9j It's true that most of the CLI linters work great, atom-hack is a CLI linter too, but some linters need methods like findFile that can only be implemented in an npm module.

AsaAyers:
It's not the new API. it's the fact that you're copying and pasting significant portions of code to manage processing and parsing.

k2b6s9j:
Ah... I see. We need to cut down the boilerplate needed for a CLI Linter.

k2b6s9j:
I have an idea. Let me implement it.

k2b6s9j:
Is it okay that my idea is service based rather than module based?

steelbrain:
:sweat_smile: I am personally against that.

AsaAyers:
As long as it gets implemented, I don't care at all how it's delivered. We have spent more time already fighting over whether it's a service or npm module than actually working on building it

k2b6s9j:
I'll put up a basic WIP PR for my idea later tonight.

k2b6s9j:
After dinner.

park9140:
:ship:

steelbrain:
Merging 'cause the next release will contain CLI helpers from #686

PR #677, opened 2015-06-30
add go to previous error

despairblue:
This is done by replacing the Set that was used to hold the errors and warnings with a double ended queue (deque). An Array could have been used but the deque has two main advantages that justify the extra dependency (which is btw tiny, 230 loc):

1. It can be cleared in a performant way
  Creating a new array would have the downsides that the commands instance would hold a reference to the old array

  Splicing the array might shrink it (who knows what v8 might eventually do, also see https://code.google.com/p/v8/issues/detail?id=2131). The deque does not shrink, which might sound awful but all references are [cleared] (https://github.com/petkaantonov/deque/blob/master/src/deque.js#L173) which allows garbage collection to kick in, also if we assume that warnings and errors appear one at a time and disappear one at a time, then shrinking the data store is quite a useless operation.

2. Handling negative indexes is way easier since this deque-implementation supports them

### DIFF
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -1,6 +1,7 @@
 BottomTab = require './views/bottom-tab'
 BottomStatus = require './views/bottom-status'
 Message = require './views/message'
+Deque = require 'double-ended-queue'
 
 class LinterViews
   constructor: (@state, @linter) ->
@@ -8,7 +9,7 @@ class LinterViews
     @showBubble = true
     @underlineIssues = true
 
-    @messages = new Set
+    @messages = new Deque
     @lineMessages = new Set
     @markers = []
     @statusTiles = []
@@ -115,11 +116,11 @@ class LinterViews
   updateBubble: (point) ->
     @removeBubble()
     return unless @showBubble
-    return unless @messages.size
+    return unless @messages.length
     activeEditor = atom.workspace.getActiveTextEditor()
     return unless activeEditor?.getPath()
     point = point || activeEditor.getCursorBufferPosition()
-    try @messages.forEach (message) =>
+    try @messages.toArray().forEach (message) =>
       return unless message.currentFile
       return unless message.range?.containsPoint point
       @bubble = activeEditor.markBufferRange([point, point], {invalidate: 'never'})
@@ -138,7 +139,7 @@ class LinterViews
     @lineMessages.clear()
     currentLine = atom.workspace.getActiveTextEditor()?.getCursorBufferPosition()?.row
     if currentLine
-      @messages.forEach (message) =>
+      @messages.toArray.forEach (message) =>
         if message.currentFile and message.range?.intersectsRow currentLine
           @lineMessages.add message
       @tabs['Line'].count = @lineMessages.size
@@ -188,7 +189,7 @@ class LinterViews
   renderPanelMarkers: ->
     @removeMarkers()
     activeEditor = atom.workspace.getActiveTextEditor()
-    @messages.forEach (message) =>
+    @messages.toArray().forEach (message) =>
       return if @state.scope isnt 'Project' and not message.currentFile
       if message.currentFile and message.range #Add the decorations to the current TextEditor
         @markers.push marker = activeEditor.markBufferRange message.range, {invalidate: 'never'}
@@ -208,7 +209,7 @@ class LinterViews
     return @setPanelVisibility(false) unless messages.size
     @setPanelVisibility(true)
     @panel.innerHTML = ''
-    messages.forEach (message) =>
+    messages.toArray.forEach (message) =>
       return if @state.scope isnt 'Project' and not message.currentFile
       Element = Message.fromMessage(message, addPath: @state.scope is 'Project', cloneNode: true)
       @panel.appendChild Element
@@ -235,7 +236,7 @@ class LinterViews
         else
           counts.project++
           message.currentFile = false
-        @messages.add message
+        @messages.push message
 
   # this method is called on package deactivate
   destroy: ->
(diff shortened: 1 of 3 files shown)


steelbrain:
The implementation looks nice, except that it adds a new dep. and conflicts with https://github.com/AtomLinter/Linter/pull/664

lib/linter-views.coffee
```diff
@@ -115,11 +116,11 @@ class LinterViews
   updateBubble: (point) ->
     @removeBubble()
     return unless @showBubble
-    return unless @messages.size
+    return unless @messages.length
     activeEditor = atom.workspace.getActiveTextEditor()
     return unless activeEditor?.getPath()
     point = point || activeEditor.getCursorBufferPosition()
-    try @messages.forEach (message) =>
+    try @messages.toArray().forEach (message) =>
```
despairblue: these are the only things I don't really like, they could be replaced with comprehensions:
```coffee
try
  for index in [0..@messages.length - 1]
    message = @messages.get(index)
    continue unless message.currentFile
    continue unless message.range?.containsPoint point
    @bubble = activeEditor.markBufferRange([point, point], {invalidate: 'never'})
    activeEditor.decorateMarker(
      @bubble
      {
        type: 'overlay',
        position: 'tail',
        item: @renderBubble(message)
      }
    )
    throw null
```

That would be faster since otherwise *deque* would always create a new array in `toArray`.
Another option would be to monkey patch *deque* to support `forEach`.

despairblue:
True about the dependency, but I fear I would reimplement a lot of *deque's* functionality just to keep it fast.

I'll have a look at your PR and try to rebase this one on it...tomorrow :sleeping:

AsaAyers:
Where do you compare against where the cursor is so that you go to the next/previous message?

I'm against the new dependency because it's a premature micro-optimization. We're not talking about a game where things happening constantly, this is a linter that triggers when the user stops typing, or you need to go to the next error when the user presses a button.

despairblue:
You are calling `@messages.add` just about every keystroke.
It can be done with arrays, as mentioned, it's just gonna be slower and look less readable, unless one abstracts the array away, but I fear that will just lead to reinventing the wheel (just slower, since I think that petkaantonov knows what he's doing)

But in the end it's your call.

Also I would have to look at @steelbrain's changes to see if my views still apply.

AsaAyers:
> You are calling @messages.add just about every keystroke.

If this is true, I'm going to need to find a new linter. That would be a really stupid thing to do

steelbrain:
@despairblue We are not calling `@messages.add` on every keystroke, we are calling it whenever the user stops typing and has onFly linters, otherwise we're only calling it on save.

AsaAyers:
http://ericleads.com/2013/04/youre-optimizing-the-wrong-things/

> Know what to optimize. If you work too hard for a 1% gain, you miss 99% of your potential.
> 
> Concentrate on the big time wasters, first.

The specifics there don't all apply to Atom development, but the core does. These kinds of micro-optimizations are not where we should be spending time.

IMO you also need to remember that the purpose of the code is to communicate with future developers (including your future-self). Adding this dependency means anyone who works with it has to learn about this new thing, where if you just use an Array then we all instantly understand it. I consider anything that is a small enough gain that the user can't perceive a difference to be a micro-optimization.

despairblue:
@AsaAyers and  @steelbrain I probably struck a nerve here. I'm sorry if I seemed rude. :anguished:

First, I said *just about*. Yeah, It is debounced, but still it lints my files with *just about* every keystroke since I normally type a couple of letters and make a short break to think.

Second, the number of operations shouldn't be underestimated. If one goes from one coding style to another (like in a project I'm workin on) it's easy to have 50 or more errors/warnings. That is noticable, though probably not because of the add operations alone.

In the end performance is not what this PR is about and I used deque because of code readability since Arrays cannot work with negative indexes which leads to a lot hard to read math code or if statements (thus implementation might also take longer if using and array) and `@messages.splice(0, @messages.length)` or `@messages.length = 0` does not convey the same meaning as `@messages.clear()`.

But as said above, I can use an array if wished for. And it seems that's the case. :smiley: 

### off topic
@AsaAyers Also I'm in favor of domain models and abstraction because of the code's purpose to communicate. The kind of optimizations you mentioned are more like the ones you would find inside of something like deque, bluebird of imutablejs, but they are abstracted away behind a well documented api that made stuff I wanted to do easier and the code more readable and saved time. :wink:
I could also make the code more readable by creating methods like `clearArray` or `getMessageFromArray` for negative indexes, but why, somebody else did that already. :expressionless:

steelbrain:
@despairblue what If I told you that the PR I mentioned above makes `@messages` an array instead of a set, so you can just store the index somewhere, and `--` it to get the previous error.

despairblue:
I will have a look at it, right now I'm just trying to go to bed, I'm close to being 24h awake, which doesn't help at all ^^

despairblue:
superseded by #678

AsaAyers:
If we have a performance problem when we have 50 messages in a file we need
to address that specifically. I don't know if it might be better to attempt
to diff the messages. I suspect for on the fly listing were rarely adding
or removing more than 2 errors per pass.

I think we should have a dummy linter for load testing. Maybe it should
complain about each character in the file.

despairblue:
@AsaAyers could be worth having a some kind of benchmark, because it's easier to reason about data, so the dummy linter sounds like a good idea.

Here are some profiles for poorly formatted a 160 loc file with about 160 warnings.

screenshot from 2015-07-01 11-31-21
screenshot from 2015-07-01 11-29-27


Also you're right in the sense that the array's push operation is probably not the culprit here. And @steelbrain's decision to get rid of getters and reinitialize the `@messages` array every time, together with coffeescripts dividend dependent modulo operator made the code in #678 quite readable. :smiley:

PR #678, opened 2015-07-01
add go to previous error

despairblue:


lib/commands.coffee
```diff
@@ -51,12 +52,26 @@ class Commands
       atom.notifications.addError error.message, {detail: error.stack, dismissable: true}
 
   nextError: ->
-    if not @messages or (next = @messages.next()).done
-      next = (@messages = @linter.views.getMessages().values()).next()
-    return if next.done # There's no errors
-    message = next.value
-    return unless message.filePath
-    return unless message.range
+    if @index?
```
park9140: Seems like we should really be indexing errors by some sort of hash of position and error message.  Otherwise changes to the full set of messages would not have a stable order.  This is of course assuming our linter messages have a stable order.

On that note we should really have a stable order, because it doesn't look like multiple linters actually have a stable order in our error processing.

lib/commands.coffee
```diff
@@ -51,12 +52,26 @@ class Commands
...
+      @index++
+    else
+      @index = 0
```
park9140: starting from 0 seems incorrect when a user cursor position is nearer to a specific error.

lib/commands.coffee
despairblue: True. I tried it with one linter, that was stable, but if there are tow and they report asynchronously it's going to be a problem as soon as one changes something.

The behavior before this PR was to always jump to the first reported error after an edit.

lib/commands.coffee
```diff
@@ -51,12 +52,26 @@ class Commands
...
+    messages = @linter.views.messages
+    message = messages[@index %% messages.length]
```
park9140: I find this kind of magic to be less than intuitive.  When you do something like this please enclose it in a method to explain what it does for those reading the code.  

Also just to make sure I interpreted this correctly.  You are basically taking the mod of the length so that just increasing the index value past the end will just loop around to the beginning. 

Maybe have a method called getCurrentMessageIndex()?

lib/commands.coffee
despairblue: Yep, but I don't like the idea to iterate over all errors to find the next closest one. :worried:

park9140:
:ship:  Seems fine other than the overarching issue of usability.  We should really add a story to address the order of errors across linter.

lib/commands.coffee
park9140: Like the order problem this will probably require a more structured message list to operate efficiently.  Something that knows which file something is in and when inserted does an ordered insert based on error start position.

lib/commands.coffee
despairblue: Yes, it is also taking the dividend independent modulo so that negative indexes work as well, taken from the coffeescript site:
```coffee
-7 %% 5 == 3 # n %% 5 is always between 0 and 4
```

I could add a comment, I could also add a function, but it seems weird to me to increase the `@index` in this method but then to call another method to use the index.

How about an `getCurrentMessage(index)` method?

lib/commands.coffee
despairblue: Yes and some api like that can take a cursor and return the next and previous error from that structure (excluding the one the cursor it at right now)

lib/commands.coffee
despairblue: Well, I mean `getMessage(index)` or `getCurrentMessage()`.

despairblue:
I extracted the modulo into a method and added an explanation.

steelbrain:
Merging based on review from @park9140

PR #686, opened 2015-07-02
Create Helpers for Easy CLI Linter Creation.

k2b6s9j:
I'm not done with this yet. But it's time for this to have some exposure.

I've created a `linter-ruby` example, which I might just merge into `linter-ruby` later.
```coffee
module.exports = LinterRuby =
  config:
    rubyExecutablePath:
      default: 'ruby'
      type: 'string'

  activate: ->
    # Show the user an error if they do not have an appropriate linter based
    #   package installed from Atom Package Manager. This will not be an issue
    #   after a base linter package is integrated into Atom, in the coming
    #   months.
    # TODO: Remove when Linter Base is integrated into Atom.
    atom.notifications.addError(
      'Linter package not found.',
      {
        detail: 'Please install the `linter` package in your Settings view'
      }
    ) unless atom.packages.getLoadedPackages 'linter'

  provideLinter: ->
    regex = '(?<file>.+):(?<line>\\d+):\\s(?<type>(?:\\w+(?:\\s)?)(?:error|warning))(?:,\\s)?(?<message>.+)'
    command = "#{atom.config.get('linter-ruby.rubyExecutablePath')} -wc"
    provider =
      scope: 'file'
      lintOnFly: false
      grammarScopes: ['source.ruby', 'source.ruby.rails', 'source.ruby.rspec']
      lint: (activeEditor, helpers) ->
        file = activeEditor.getPath()
        helpers.execFilePath(command, file, {stream: 'stderr'}).then (output) ->
          helpers.parse(output, regex)
```

It works:
screen shot 2015-07-02 at 3 42 21 pm

Should be merged in before #676

### DIFF
--- a/lib/helpers.coffee
+++ b/lib/helpers.coffee
@@ -40,6 +40,6 @@ module.exports = Helpers =
     throw new Error "No File Path to work with." if not filePath
     path = require 'path'
     return new Promise (resolve, reject) ->
-      file = path.baseName(filePath)
-      options.cwd = path.baseDir(filePath) if not options.cwd
-      resolve(exec("#{command} #{file}"), options)
+      file = path.basename(filePath)
+      options.cwd = path.dirname(filePath) if not options.cwd
+      resolve(Helpers.exec("#{command} #{file}"), options)
--- a/lib/helpers.coffee
+++ b/lib/helpers.coffee
@@ -35,11 +35,13 @@ module.exports = Helpers =
     throw new Error "Nothing to execute." if not arguments.length
     return new Promise (resolve, reject) ->
       process = child_process.exec(command, options)
+      options.stream = 'stdout' if not options.stream
       data = []
       if options.stream == 'stderr'
         process.stderr.on 'data', (d) -> data.push(d.toString())
       else
         process.stdout.on 'data', (d) -> data.push(d.toString())
+      process.stdin.write(options.stdin.toString()) if options.stdin
       process.on 'close', ->
         resolve(data)
 
@@ -75,36 +77,27 @@ module.exports = Helpers =
       toReturn = []
       regex = XRegExp(regex)
       for line in data
-        match = XRegExp.exec(line, regex)
-        if match
-          if match.lineStart
-            lineStart = match.lineStart - options.baseReduction
-          else if match.line
-            lineStart = match.line - options.baseReduction
-          else
-            lineStart = 0
-          if match.colStart
-            colStart = match.colStart - options.baseReduction
-          else if match.col
-            colStart = match.col - options.baseReduction
-          else
-            colStart = 0
-          if match.lineEnd
-            lineEnd = match.lineEnd - options.baseReduction
-          else if match.line
-            lineEnd = match.line - options.baseReduction
-          else
-            lineEnd = 0
-          if match.colEnd
-            colEnd = match.colEnd - options.baseReduction
-          else if match.col
-            colEnd = match.col - options.baseReduction
-          else
-            colEnd = 0
+        if regex.test line
+          match = XRegExp.exec(line, regex)
+          options.baseReduction = 1 if not options.baseReduction
+          lineStart = 0
+          lineStart = match.line - options.baseReduction if match.line
+          lineStart = match.lineStart - options.baseReduction if match.lineStart
+          colStart = 0
+          colStart = match.col - options.baseReduction if match.col
+          colStart = match.colStart - options.baseReduction if match.colStart
+          lineEnd = 0
+          lineEnd = match.line - options.baseReduction if match.line
+          lineEnd = match.lineEnd - options.baseReduction if match.lineEnd
+          colEnd = 0
+          colEnd = match.col - options.baseReduction if match.col
+          colEnd = match.colEnd - options.baseReduction if match.colEnd
+          filePath = match.file
+          filePath = options.filePath if options.filePath
           toReturn.push(
             type: match.type,
             text: match.message,
-            filePath: match.file
+            filePath: filePath,
             range: [[lineStart, colStart], [lineEnd, colEnd]]
           )
       resolve(toReturn)
--- a/lib/helpers.coffee
+++ b/lib/helpers.coffee
@@ -27,11 +27,19 @@ module.exports = Helpers =
   # Everything past this point relates to CLI helpers as loosly demoed out in:
   #   https://gist.github.com/steelbrain/43d9c38208bf9f2964ab
 
-  exec: (command, options = {}) ->
+  exec: (command, options = {stream: 'stdout'}) ->
     throw new Error "Nothing to execute." if not arguments.length
     child_process = require 'child_process'
     return new Promise (resolve, reject) ->
-      resolve(child_process.exec(command, options))
+      process = child_process.exec(command, options)
+      data = []
+      if options.stream == 'stderr'
+        process.stderr.on 'data', (d) -> data.push(d.toString())
+      else
+        process.stdout.on 'data', (d) -> data.push(d.toString())
+      process.on 'close', ->
+        console.log data
+        resolve(data)
 
   # This should only be used if the linter is only working with files in their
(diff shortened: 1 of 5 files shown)


k2b6s9j:
Some more tests should be written as well.

steelbrain:
:clap:

k2b6s9j:
@steelbrain Was a great learning experience. I know more about Asynchronous JavaScript/CoffeeScript than ever before.

steelbrain:
Great!

lib/helpers.coffee
```diff
@@ -1,27 +1,94 @@
...
+      regex = XRegExp(regex)
+      for line in data
+        match = XRegExp.exec(line, regex)
+        console.log match
+        if match
+          toReturn.push(
+            type: match.type,
+            text: match.message,
+            filePath: match.file
+            range: [[match.line - 1, 0], [match.line - 1, 0]]
```
steelbrain: instead of setting end range to zero, why not check the length of the row since you already have `TextEditor` and set it according to that.
Also this one is tailored for `linter-ruby`, it should also allow others that specify a full range like `startLine`, `startCol`, `endLine` and `endCol`

k2b6s9j:
Part of the problem with this implementation is that it doesn't currently do linting on the fly. Which `linter-ruby` did before. I could use some help creating an on-the-fly linting helper.

steelbrain:
What CLI linter does ruby use btw?

k2b6s9j:
MRI comes with some basic checking built in using a parameter, `ruby -wc [FILENAME]`

lib/helpers.coffee
k2b6s9j: Not finished, yet. That is part of my fine tuning. I really just wanted to show *something* for now.

lib/helpers.coffee
```diff
@@ -1,27 +1,94 @@
...
+        console.log data
+        resolve(data)
+
+  # This should only be used if the linter is only working with files in their
+  #   base directory. Else wise they should use `Helpers#exec`.
+  execFilePath: (command, filePath, options = {}) ->
+    throw new Error "Nothing to execute." if not arguments.length
+    throw new Error "No File Path to work with." if not filePath
+    path = require 'path'
+    child_process = require 'child_process'
```
steelbrain: how about you make this require global?

lib/helpers.coffee
```diff
@@ -1,27 +1,94 @@
...
+      process.on 'close', ->
+        console.log data
+        resolve(data)
+
+  # This should only be used if the linter is only working with files in their
+  #   base directory. Else wise they should use `Helpers#exec`.
+  execFilePath: (command, filePath, options = {}) ->
+    throw new Error "Nothing to execute." if not arguments.length
+    throw new Error "No File Path to work with." if not filePath
+    path = require 'path'
```
steelbrain: and this one too

k2b6s9j: Can do.

steelbrain:
If it accepts output via stdin, You could just do
```js
@execFilePath(command, filePath, {stream: stderr, stdin: textEditor.getText()})
```
and pass that into child_process methods

lib/helpers.coffee
k2b6s9j: Can do.

lib/helpers.coffee
steelbrain: and You did a great job in that manner. :clap:

k2b6s9j:
@steelbrain I did the changes to line and column logic. But I didn't do the auto detection of the line ending like you were saying. In my experience when I had tried to do that I would always find some issue with the file results being returned for wasn't the file the TextEditor is representing.

k2b6s9j:
@steelbrain I'll experiment with stdin.

k2b6s9j:
This should be ready to merge once I implement on-the-fly linting.

steelbrain:
:tada:

k2b6s9j:
Alright I think I'm done, so I guess it's time for a final code review and request. /cc @AtomLinter/linter

lib/helpers.coffee
```diff
@@ -11,17 +16,88 @@ Helpers = module.exports =
...
+  #   base directory. Else wise they should use `Helpers#exec`.
+  execFilePath: (command, filePath, options = {}) ->
+    throw new Error "Nothing to execute." if not arguments.length
+    throw new Error "No File Path to work with." if not filePath
+    return new Promise (resolve, reject) ->
+      options.cwd = path.dirname(filePath) if not options.cwd
+      command = "#{command} #{filePath}"
+      resolve(Helpers.exec(command, options))
+
+  # Due to what we are attempting to do, the only viable solution right now is
```
steelbrain: Do we really need all this description? :stuck_out_tongue:

spec/helpers-spec.coffee
```diff
@@ -36,3 +36,16 @@ describe "The Linter Validation Helper", ->
...
+    waitsForPromise ->
+      @command = Helpers.exec('echo')
+  it "should throw when no parameters are passed.", ->
+    expect( -> Helpers.exec()).toThrow()
+
+describe "The Command and FilePath Helper", ->
+  it "should throw when no parametes are passed.", ->
+    expect( -> Helpers.execFilePath()).toThrow()
+  it "should throw when no File Path is passed.", ->
+    expect( -> Helpers.execFilePath('echo')).toThrow()
```
steelbrain: Can you do some example `exec` with the echo command?
and can you also add tests for parse by having some example input and a regex and matching their values?

lib/helpers.coffee
k2b6s9j: I see nothing wrong with some clarifying comments.

spec/helpers-spec.coffee
k2b6s9j: Yeah. It was giving me trouble earlier, but I can give it another go now. I'll probably get it to work this time.

lib/helpers.coffee
steelbrain: It's upto you, but IMO having a quick look at the code below automatically describes all of that.

steelbrain:
@k2b6s9j I think this PR has some merge conflicts in CHANGELOG.md

AsaAyers:
I see arguments aren't passed as an array. So we're relying on each linter to figure out the exact way to escape parameters correctly?

k2b6s9j:
Doh! I'll fix that.

k2b6s9j:
@steelbrain Just fixed the change log conflict by merging master in.

steelbrain:
Great, now all we need is just a few more tests.

lib/helpers.coffee
```diff
@@ -11,17 +16,88 @@ Helpers = module.exports =
...
-
     return true
+
+  # Everything past this point relates to CLI helpers as loosly demoed out in:
+  #   https://gist.github.com/steelbrain/43d9c38208bf9f2964ab
+
+  exec: (command, options = {stream: 'stdout'}) ->
+    throw new Error "Nothing to execute." if not arguments.length
+    return new Promise (resolve, reject) ->
+      process = child_process.exec(command, options)
```
AsaAyers: Don't use `exec`, use [`spawn`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) instead. With `exec` it forces you to combine command and arguments into one big string that is then going to get split and you have to use special handling to make spaces work correctly. `spawn` takes a command and an array of arguments.

k2b6s9j:
Everything has been tested. I've follow the advice of both @steelbrain and @AsaAyers. @AtomLinter/linter review and merge?

steelbrain:
LGTM

iam4x:
:tada:

steelbrain:
I've got write access to the `atom-linter` npm module, should I migrate these there?
/cc @AtomLinter/linter

AsaAyers:
I thought that's where they were already. Seems like the right place to me.

steelbrain:
@AsaAyers https://github.com/AtomLinter/Linter/pull/694

PR #689, opened 2015-07-03
prevent badge overflow

braver:
Badges in the tooltips overflow outside their colored box. This at least prevents that, but perhaps it could be improved further by making sure the badge grows with its content.

before

after


iam4x:
Thanks a lot!

Issue #687, opened 2015-07-03
XRegExp is just too heavy


steelbrain:
We are using it solely for simplicity. While it makes writing providers easier, it eats a lot of CPU and we're trying to avoid things that do that.
The only proposed fix I have in mind is that we create our own lightweight alternative that works with the same syntax and translates that regex to native, and then remaps those indices to names extracted before. and most of all, caches the compiled Regex.
It would take only like a day or two and we already have tests in the wild for it. so I think we should take this option.

k2b6s9j:
Of course, those building their own full featured linters (<15 lines) should be using Vanilla JavaScript Regex anyway.

steelbrain:
Or maybe we could just cache existing xregexp instead of recreating it each time we wanna process messages. I'll see how that one goes.

PR #693, opened 2015-07-05
Cache MessageRegexp

steelbrain:
Fixes https://github.com/AtomLinter/Linter/issues/687

### DIFF
--- a/lib/linter.coffee
+++ b/lib/linter.coffee
@@ -2,7 +2,6 @@ fs = require 'fs'
 path = require 'path'
 {CompositeDisposable, Range, Point, BufferedProcess} = require 'atom'
 XRegExp = null
-MessageRegexp = null
 {deprecate} = require('grim')
 
 # These are NOOPs in linter-plus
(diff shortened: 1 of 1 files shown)


lib/linter.coffee
```diff
@@ -2,6 +2,7 @@ fs = require 'fs'
 path = require 'path'
 {CompositeDisposable, Range, Point, BufferedProcess} = require 'atom'
 XRegExp = null
+MessageRegexp = null
```
iam4x: Can't we keep `MessageRegexp` on the `Linter` instance?

steelbrain: We can, but the linter instance isn't shared, so I find this a bit way cleaner.

maxbrunsfeld:
I think this change caused `linter-clang` to stop working. It looks like `linter-clang` [uses a different RegExp](https://github.com/AtomLinter/linter-clang/blob/dae931d41fa8a09d4a0189202f4e5784414e924b/lib/linter-clang.coffee#L147) each time `lintFile` is called, because the RegExp is based on the filename. This caching causes the wrong RegExp to be used. I don't know if this should be considered a bug in `linter` or in `linter-clang`.

steelbrain:
In any case, It's using the legacy API, it should upgrade to solve this issue :)

AsaAyers:
@maxbrunsfeld If you'd like an actual fix that can be done immediately, it looks like you can clear the cache with this:

```CoffeeScript
    # add file to regex to filter output to this file,
    # need to change filename a bit to fit into regex
    delete @MessageRegexp
    @regex = filePath.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&") +
```

maxbrunsfeld:
Thanks @AsaAyers, @steelbrain. I've just installed linter 1.1.0 for the time being.

PR #695, opened 2015-07-05
Add linter:togglePanel command

### LINKED ISSUE #672: Request for new keybinding/command Toggle:Error Panel
It would be cool if we could toggle the error panel at the bottom. 
Sometimes I just want to hide it to have more workspace for the editor. The only way I can achieve this is by going to `settings > linter` and uncheck `Show Error Panel at the bottom` but I do not want to hide it forever. 
Can this be automated?

steelbrain:
Just like the title says
Fixes #672


steelbrain:
Minor change, self merging.

Issue #696, opened 2015-07-05
Support Linters Which Modify Buffers


steelbrain:
Continuing from https://github.com/AtomLinter/linter-plus/issues/78

steelbrain:
A fix is up in https://github.com/AtomLinter/Linter/pull/697

PR #697, opened 2015-07-05
Support buffer modifying linters

steelbrain:
Fixes https://github.com/AtomLinter/Linter/issues/696
I've confirmed that it works.
Buffer modifying linters need to have a `modifiesBuffer:true` in their object. This change is backward compatible.

### DIFF
--- a/lib/editor-linter.coffee
+++ b/lib/editor-linter.coffee
@@ -60,20 +60,29 @@ class EditorLinter
 
   # This method returns an array of promises to be used in lint
   triggerLinters: (bufferModifying, wasTriggeredOnChange, scopes) ->
-    Promises = []
+    ToReturn =
+      if bufferModifying
+        Promise.resolve()
+      else
+        []
     @linter.getLinters().forEach (linter) =>
       return unless Helpers.shouldTriggerLinter(linter, wasTriggeredOnChange, scopes)
-      Promises.push new Promise((resolve) =>
-        resolve(linter.lint(@editor, Helpers))
-      ).then((results) =>
-        if linter.scope is 'project'
-          @linter.setMessages(linter, results)
-        else
-          # Trigger event instead of updating on purpose, because
-          # we want to make MessageRegistry the central message repo
-          @emitter.emit('should-update', {linter, results})
-      ).catch (error) ->
-        atom.notifications.addError error.message, {detail: error.stack, dismissable: true}
+      currentLinter =>
+        return Promise.resolve().then( =>
+          return linter.lint(@editor, Helpers)
+        ).then((results) =>
+          if linter.scope is 'project'
+            @linter.setMessages(linter, results)
+          else
+            # Trigger event instead of updating on purpose, because
+            # we want to make MessageRegistry the central message repo
+            @emitter.emit('should-update', {linter, results})
+        ).catch (error) ->
+          atom.notifications.addError error.message, {detail: error.stack, dismissable: true}
+      if bufferModifying
+        ToReturn.then -> currentLinter()
+      else
+        ToReturn.push(currentLinter())
     Promises
 
   # This method sets or gets the lock status of given type
--- a/lib/editor-linter.coffee
+++ b/lib/editor-linter.coffee
@@ -65,8 +65,8 @@ class EditorLinter
       return if linter.modifiesBuffer isnt bufferModifying
       return unless Helpers.shouldTriggerLinter(linter, wasTriggeredOnChange, scopes)
       currentLinter = =>
-        return Promise.resolve().then( =>
-          return linter.lint(@editor, Helpers)
+        return new Promise((resolve) =>
+          resolve(linter.lint(@editor, Helpers))
         ).then((results) =>
           if linter.scope is 'project'
             @linter.setMessages(linter, results)
--- a/lib/editor-linter.coffee
+++ b/lib/editor-linter.coffee
@@ -66,7 +66,12 @@ class EditorLinter
       Promises.push new Promise((resolve) =>
         resolve(linter.lint(@editor, Helpers))
       ).then((results) =>
-        @gotLinterResults(linter, results)
+        if linter.scope is 'project'
+          @linter.setMessages(linter, results)
+        else
+          # Trigger event instead of updating on purpose, because
+          # we want to make MessageRegistry the central message repo
+          @emitter.emit('should-update', {linter, results})
       ).catch (error) ->
         atom.notifications.addError error.message, {detail: error.stack, dismissable: true}
     Promises
--- a/lib/editor-linter.coffee
+++ b/lib/editor-linter.coffee
@@ -53,45 +53,22 @@ class EditorLinter
 
     scopes = @editor.scopeDescriptorForBufferPosition(@editor.getCursorBufferPosition()).scopes
     scopes.push '*' # To allow global linters
-    @triggerBufferModifyingLinters.then( =>
-      return Promise.all(@triggerLinters(wasTriggeredOnChange, scopes))
+    @triggerLinters(true, wasTriggeredOnChange, scopes).then( =>
+      return Promise.all(@triggerLinters(false, wasTriggeredOnChange, scopes))
     ).then =>
       @lock(wasTriggeredOnChange, false)
 
-  triggerBufferModifyingLinters: (wasTriggeredOnChange, scopes)->
-    Sequence = Promise.resolve()
-    @linter.getLinters().forEach (linter) =>
-      return unless linter.modifiesBuffer
-      return unless Helpers.shouldTriggerLinter(linter, wasTriggeredOnChange, scopes)
-      Sequence = Sequence.then( =>
-        linter.lint(@editor)
-      ).then((results) =>
-        if linter.scope is 'project'
-          @linter.setMessages(linter, results)
-        else
-          # Trigger event instead of updating on purpose, because
-          # we want to make MessageRegistry the central message repo
-          @emitter.emit('should-update', {linter, results})
-      )
-    return Sequence
-
   # This method returns an array of promises to be used in lint
-  triggerLinters: (wasTriggeredOnChange, scopes) ->
+  triggerLinters: (bufferModifying, wasTriggeredOnChange, scopes) ->
     Promises = []
     @linter.getLinters().forEach (linter) =>
       return unless Helpers.shouldTriggerLinter(linter, wasTriggeredOnChange, scopes)
       Promises.push new Promise((resolve) =>
         resolve(linter.lint(@editor, Helpers))
       ).then((results) =>
-        if linter.scope is 'project'
-          @linter.setMessages(linter, results)
-        else
-          # Trigger event instead of updating on purpose, because
-          # we want to make MessageRegistry the central message repo
-          @emitter.emit('should-update', {linter, results})
+        @gotLinterResults(linter, results)
       ).catch (error) ->
         atom.notifications.addError error.message, {detail: error.stack, dismissable: true}
-
     Promises
 
   # This method sets or gets the lock status of given type
--- a/lib/editor-linter.coffee
+++ b/lib/editor-linter.coffee
@@ -60,11 +60,7 @@ class EditorLinter
 
   # This method returns an array of promises to be used in lint
   triggerLinters: (bufferModifying, wasTriggeredOnChange, scopes) ->
-    ToReturn =
-      if bufferModifying
-        Promise.resolve()
-      else
-        []
+    ToReturn = if bufferModifying then Promise.resolve() else []
     @linter.getLinters().forEach (linter) =>
       return if linter.modifiesBuffer isnt bufferModifying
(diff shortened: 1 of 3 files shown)


lib/helpers.coffee
```diff
@@ -4,6 +4,12 @@ path = require 'path'
 child_process = require 'child_process'
 
 Helpers = module.exports =
+  shouldTriggerLinter: (linter, wasTriggeredOnChange, scopes)->
+    # Trigger fly linters on save, but not save linters on fly
```
AsaAyers: If you're going to separate these, we need a new category for CoffeeLint which does both.

steelbrain: Quite confusing, would you please elaborate?

AsaAyers: Sorry, I misread the comment thinking fly linters would *not* trigger on save. It's only the reverse.

CoffeeLint will lint on the fly, but there are times (often when editing the config file) when I save specifically to trigger the linter.

lib/editor-linter.coffee
```diff
@@ -53,30 +53,34 @@ class EditorLinter
...
+    @triggerLinters(true, wasTriggeredOnChange, scopes).then( =>
+      return Promise.all(@triggerLinters(false, wasTriggeredOnChange, scopes))
+    ).then =>
       @lock(wasTriggeredOnChange, false)
 
   # This method returns an array of promises to be used in lint
-  triggerLinters: (wasTriggeredOnChange, scopes) ->
-    Promises = []
+  triggerLinters: (bufferModifying, wasTriggeredOnChange, scopes) ->
+    ToReturn = if bufferModifying then Promise.resolve() else []
```
AsaAyers: It's not a good idea to have a variable that alternates between different types. Sometimes it's a `Promise`, sometimes it's an `Array`. Off the top of my head I don't know what the exact solution is here, but I don't think you should combine these.

lib/editor-linter.coffee
```diff
@@ -53,30 +53,34 @@ class EditorLinter
...
+          if linter.scope is 'project'
+            @linter.setMessages(linter, results)
+          else
+            # Trigger event instead of updating on purpose, because
+            # we want to make MessageRegistry the central message repo
+            @emitter.emit('should-update', {linter, results})
+        ).catch (error) ->
+          atom.notifications.addError error.message, {detail: error.stack, dismissable: true}
+      if bufferModifying
+        ToReturn.then -> currentLinter()
```
AsaAyers: Do linters that can modify the buffer not get to return messages? This implementation will only keep the messages from the last linter that modified the buffer.

lib/editor-linter.coffee
```diff
@@ -53,30 +53,34 @@ class EditorLinter
...
+      currentLinter = =>
+        return new Promise((resolve) =>
+          resolve(linter.lint(@editor, Helpers))
+        ).then((results) =>
+          if linter.scope is 'project'
+            @linter.setMessages(linter, results)
+          else
+            # Trigger event instead of updating on purpose, because
+            # we want to make MessageRegistry the central message repo
+            @emitter.emit('should-update', {linter, results})
```
AsaAyers: Is `MessageRegistry` in place? Shouldn't we set the value and have it emit to notify the views that they need to update?

I think maybe this is an indirect setter, and I've found that strategy to not work very well as the system grows.

lib/editor-linter.coffee
```diff
@@ -53,30 +53,34 @@ class EditorLinter
...
+      return Promise.all(@triggerLinters(false, wasTriggeredOnChange, scopes))
+    ).then =>
       @lock(wasTriggeredOnChange, false)
 
   # This method returns an array of promises to be used in lint
-  triggerLinters: (wasTriggeredOnChange, scopes) ->
-    Promises = []
+  triggerLinters: (bufferModifying, wasTriggeredOnChange, scopes) ->
+    ToReturn = if bufferModifying then Promise.resolve() else []
     @linter.getLinters().forEach (linter) =>
```
AsaAyers: does `getLinters()` return an Array? Right now this whole `ToReturn` thing is really just simulating a `.map()` or a `.reduce()` depending on if the linters modify the buffer. If you can I recommend `return @linter.getLinters().reduce(...)` instead

park9140:
We really need unit tests before we start making changes like this.

park9140:
Can you explain the full process you are using for handling buffer modifying linters?  I think there are things missing here but I can't really follow your code to point them out in any exact manner.

lib/editor-linter.coffee
steelbrain: Not really, it'll keep the messages from all linters, and not only that, It'll only run the bufferModifying linters in series, not in parallel.

lib/editor-linter.coffee
steelbrain: While I agree with the approach, I had two options, complexity or this, I chose this. You can find more about how this works in the onShouldUpdate observer in messaeg-registry

lib/editor-linter.coffee
steelbrain: `getLinters()` doesn't return an array, It's an ES6 map. I was initially gonna do map/reduce but I figured this approach to be simpler (I tried it first then deleted the code)

steelbrain:
@park9140 

1. Lint method is triggered
2. We trigger all the bufferModifying linters
  1. Non-Buffer modifying linters are skipped
  2. Common logic is stored in a module
  3. a .then handler is added to the `ToReturn` and the return value of common logic function is returned from there
  4. The last promise resolves when all the promises have, in sequence
3. We trigger the non-buffer-modifying linters as usual

lib/editor-linter.coffee
AsaAyers: I think you're wrong. A test would prove it one way or the other.

steelbrain:
I think I am gonna add some tests for it

lib/editor-linter.coffee
AsaAyers: Oh, I see. It works because you're relying on  side effects. The return value doesn't actually matter. If it did this would fail.

steelbrain: I still wouldn't call it a side-effect though

park9140:
@steelbrain, how does this handle the situation of multiple buffer modifying linters?  Do the buffer modifying linters get run in series?

steelbrain:
They run in sequence, not in parallel, so only one buffer modifying linter would be executed at a time, so it makes it safe to have multiple buffer modifying linters.

steelbrain:
@k2b6s9j your spec about a linter error is logging an error in the console but still failing

steelbrain:
The tests have been pushed people. They are all passing.

spec/buffer-modifying-provider-spec.coffee
```diff
@@ -0,0 +1,77 @@
...
+        expect(activeEditor.getText()).toBe('second')
+        activeEditor.setText('third')
+        return []
+    normalLinter =
+      grammarScopes: ['*']
+      scope: 'file'
+      lintOnFly: false
+      modifiesBuffer: false
+      lint: ->
+        expect(activeEditor.getText()).toBe('third')
```
park9140: expectations should not be part setup, this makes tests hard to follow.  store the information then write the expectations below.

spec/buffer-modifying-provider-spec.coffee
```diff
@@ -0,0 +1,77 @@
...
+      modifiesBuffer: false
+      lint: ->
+        last = 'normal'
+        return []
+    linter.addLinter(bufferModifying)
+    linter.addLinter(normalLinter)
+    waitsForPromise ->
+      linter.getActiveEditorLinter().lint(false).then ->
+        expect(last).toBe('normal')
+  it 'runs in parallel', ->
```
park9140: 'runs buffer modifying linters in series by registration order'?

spec/buffer-modifying-provider-spec.coffee
steelbrain: it's not really a setup, it demos how first linter set's it's text to first and second linter verifies it and second linter then sets a text and third verifies it. I think that was the best way to test that it runs in a chain.

spec/buffer-modifying-provider-spec.coffee
steelbrain: already fixed, refresh your page.

spec/buffer-modifying-provider-spec.coffee
park9140: Everything that does execution of code is setup.

spec/buffer-modifying-provider-spec.coffee
steelbrain: It was a typo, it should've been `run in sequence`

spec/buffer-modifying-provider-spec.coffee
steelbrain: :sweat_smile: I don't follow, example please?

spec/buffer-modifying-provider-spec.coffee
```diff
@@ -0,0 +1,77 @@
...
+      modifiesBuffer: false
+      lint: ->
+        last = 'normal'
+        return []
+    linter.addLinter(bufferModifying)
+    linter.addLinter(normalLinter)
+    waitsForPromise ->
+      linter.getActiveEditorLinter().lint(false).then ->
+        expect(last).toBe('normal')
+  it 'run in sequence', ->
```
park9140: `it runs buffer modifying linters in series`

spec/buffer-modifying-provider-spec.coffee
park9140: Also this has to many expectations in one test.  You actually have a couple tests going on here that could run off of the same setup.

spec/buffer-modifying-provider-spec.coffee
steelbrain: we are actually describing `buffer modifying linters` above, so we don't have to write that whole message, when we write `run in sequence` it becomes `buffer modifying linters run in sequence`

spec/buffer-modifying-provider-spec.coffee
```diff
@@ -0,0 +1,77 @@
+describe 'buffer modifying linters', ->
+  getModuleMain = -> atom.packages.getActivePackage('linter').mainModule.instance
+  beforeEach ->
+    waitsForPromise ->
+      atom.packages.activatePackage('linter')
+    waitsForPromise ->
+      atom.workspace.open('test.txt')
+  it 'are triggered before other linters', ->
```
park9140: `it 'triggers buffer modifying linters first'`

steelbrain: done

spec/buffer-modifying-provider-spec.coffee
steelbrain: You mean write a function that retuns a provider and then use it in tests instead of writing it each time?

park9140: nest a describe for the sequence testing.  Do the setup portions of this then write the individual tests that exist here.
```coffeescript
describe 'order of execution of buffer modifying linters', ->
  beforeEach ->
    // set up all the linters here, copy the editor value when each linter was executed into a variable for that linter
  it 'should execute the linters in order of registration` ->
    expect(secondLinterEditorValue).toBe('first')
  it 'should execute async linters in correct order'
    expect(asyncLinterEditorValue).toBe('second')
  it 'should execute non modifying linters last' ->
    expect(nonModifyingEditorValue).toBe('third')
```

Additionally you need to have timeouts and

park9140: More appropriately you could just push the linter onto an array as it is executed then test the position in the array is correct rather than testing the actual editor values...

steelbrain: makes sense, thanks!

steelbrain: I am a bit confused, we don't really handle the triggering of the linters ourselves, we add them to as linters, linter does the rest.
Also pushing the values to an array and then verifying their indices won't be very different from what's already implemented, do you have an alternative approach in mind?

park9140: I mean when you define the linters for the test.  
```
linterExecutionOrder = []
third =
      grammarScopes: ['*']
      scope: 'file'
      lintOnFly: false
      modifiesBuffer: true
      lint: ->
        linterExecutionOrder.push(this)
        activeEditor.setText('third')
        return []

expect(linterExecutionOrder[3]).toBe(third);

steelbrain: done

steelbrain:
Any further suggestions?

steelbrain:
Merging in on :+1: from @AsaAyers and because there were no known unresolved bugs.

iam4x:
Woot! We will be able to include https://atom.io/packages/jscs-fixer directly into `linter-jscs` if I'm right!

:clap:

steelbrain:
You are :tada:

PR #712, opened 2015-07-08
Indicate when one or more lint job is running

b52:
The bottom bar is updated to show if at least one lint job is running. This is very helpful with larger projects where lint jobs might take some time. This is just my idea how one could visualize this state.

Since this is my first contact with Linter, I would argue that you should very extensive check that everything is in order.

### DIFF
--- a/lib/linter-plus.coffee
+++ b/lib/linter-plus.coffee
@@ -20,6 +20,7 @@ class Linter
     @emitter = new Emitter
     @editorLinters = new Map
     @linters = new Set # Values are pushed here from Main::consumeLinter
+    @lintJobs = 0
 
     @messages = new Messages @
     @views = new LinterViews @
@@ -32,6 +33,9 @@ class Linter
 
     @subscriptions.add atom.workspace.observeTextEditors (editor) =>
       currentEditorLinter = new EditorLinter @, editor
+      currentEditorLinter.onLinting (linting) =>
+        @lintJobs += if linting then 1 else -1
+        @emitter.emit 'linting', @lintJobs > 0
       @editorLinters.set editor, currentEditorLinter
       @emitter.emit 'observe-editor-linters', currentEditorLinter
       currentEditorLinter.lint false
@@ -70,6 +74,9 @@ class Linter
 
   getMessages: ->
     return @messages.getAll()
+
+  onLinting: (callback) ->
+    @emitter.on 'linting', callback
 
   onDidChangeMessages: (callback) ->
     return @messages.onDidChange(callback)
--- a/lib/linter-plus.coffee
+++ b/lib/linter-plus.coffee
@@ -34,7 +34,7 @@ class Linter
     @subscriptions.add atom.workspace.observeTextEditors (editor) =>
       currentEditorLinter = new EditorLinter @, editor
       currentEditorLinter.onLinting (linting) =>
-        @lintJobs += if linting then 1 else -1
+        if linting then @lintJobs++ else @lintJobs--
         @emitter.emit 'linting', @lintJobs > 0
       @editorLinters.set editor, currentEditorLinter
       @emitter.emit 'observe-editor-linters', currentEditorLinter
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -2,7 +2,6 @@
 
 BottomPanel = require('./views/bottom-panel')
 BottomContainer = require('./views/bottom-container')
-BottomStatus = require('./views/bottom-status')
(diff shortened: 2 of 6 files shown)


lib/linter-plus.coffee
```diff
@@ -32,6 +33,9 @@ class Linter
 
     @subscriptions.add atom.workspace.observeTextEditors (editor) =>
       currentEditorLinter = new EditorLinter @, editor
+      currentEditorLinter.onLinting (linting) =>
+        @lintJobs += if linting then 1 else -1
```
iam4x: `if linting then @lintsJobs++ else @lintsJobs--`

But I think it's just preference of how to write it. (I find it clearer)

lib/views/bottom-container.coffee
```diff
@@ -1,4 +1,3 @@
-
```
iam4x: I love that <3

lib/views/bottom-status.js
```diff
@@ -9,24 +9,50 @@ class BottomStatus extends HTMLElement{
     this.iconSpan.classList.add('icon')
     this.appendChild(this.iconSpan)
 
-    this.count = 0
+    this._linting = false
+    this._count = 0
```
iam4x: We tried to remove all the `_method / _var` to please write it `this.linting` etc...

lib/views/bottom-container.coffee
steelbrain: Hahaha

styles/linter-plus.less
```diff
@@ -174,7 +174,10 @@ atom-text-editor::shadow .linter-highlight, .linter-highlight{
   &.status-error {
     color: @background-color-error;
   }
-  &.status-success{
+  &.status-success {
```
iam4x: :+1:

lib/linter-views.coffee
```diff
@@ -2,7 +2,6 @@
 
 BottomPanel = require('./views/bottom-panel')
 BottomContainer = require('./views/bottom-container')
-BottomStatus = require('./views/bottom-status')
```
iam4x: `BottomStatus` isn't used?

b52: No, it's only used by `BottomContainer`, which makes sense.

lib/views/bottom-status.js
steelbrain: @iam4x I think `_linting` makes sense here because we don't wanna trigger render.

lib/linter-views.coffee
steelbrain: Yep, it's an import that I forgot to remove in the recent `Compact Message Storage` PR

lib/views/bottom-status.js
iam4x: We should keep only one codestyle, if we removed all the underscores it makes sense to removes these too.

lib/linter-plus.coffee
b52: I find yours cleaner as well. It's just too late :sleeping:

lib/linter-views.coffee
iam4x: Good catch so ;)

lib/views/bottom-status.js
steelbrain: Underscore is only a place holder here, we have two options here

1. Remove getter/setters and underscores and use property names as keys and have setters and getters like `setSomething` and `getSomething`
2. Or we keep underscore for those who really need it, and use es6 setters/getters

b52:
Added the proposed simplification and removed the fields prefixed with an underscore.

steelbrain:
while this PR looks good, you are not actually counting active linters, You are counting active editors that have linting in progress.

b52:
@steelbrain You're right. What's wrong about that?

steelbrain:
@b52 the users will expect to see a number of active linters instead of active editor linters IMO

steelbrain:
Showing linters instead of editor linters there should be a fairly simple change, all you have to do is increment or decrement in `EditorLinter::triggerLinters`'s foreach loop.

b52:
@steelbrain: Pardon? I don't show any linters, thus the actual number doesn't matter. It's just important to check if any linter is running, and if so, show it. That is what I do, since I use the `Promise::all` as source applied on your `EditorLinter::triggerLinters` outcome, i.e., the array or linter promises.

So, what is left for me to do?

PS: I just made the sync icon rotate to illustrate even more that something is happening.

steelbrain:
LGTM. /cc @AtomLinter/linter

PR #731, opened 2015-07-12
[experiment] Don't merge: extracting triggerLinters into distinct pieces

AsaAyers:
`triggerLinters` was actually doing several things and seemed hard to follow. I consider this an early draft of what it might look like to break it into pieces that each do one thing.

Now there is a distinct stage for gathering all of the data, and then a stage to store that data. I'm out of time for today, but each of these functions can be tested easier than the whole `triggerLinters`


steelbrain:
Closing, as this issue was addressed by the recent LinterRegistry class.

Issue #708, opened 2015-07-07
Linter destroys performance on large files


toomuchpete:
Had to disable linter (and probably go back to Sublime Text) because Linter is causing the editor to take 1-2s to respond to keystrokes moving through a file. It's an 9,000 line file with a bunch of linting errors (1,700+) in it. 

I'm running three individual linters (`linter-php`, `linter-phpcs`, and `linter-phpmd`) if any of them are turned on, the app grinds to a halt. Seems like there might be a problem with the way the larger linter plugin handles and displays large number of linting errors.

steelbrain:
You have `1700` errors in one file, Woot! :sweat_smile: 
Rest assured we are working on making linter more efficient and faster, we already have a lot of performance improvements in the master branch and will be released soon.
Until then, You can just disable linting for that one particular file while still using it all over your project, by focus that text editor and running `linter:toggle` from the command pallete.

toomuchpete:
@steelbrain Thanks for the quick reply! I'll keep an eye on future releases, but not having access to it for the bigger files is pretty much a deal breaker for me right now, so back to ST I go.

steelbrain:
While profiling today, I set up a few hooks in the DOM updates and realized that we are doing stuff really inefficiently, I have some updates planned that should fix them.

nemoDreamer:
:+1: could we have a setting for the throttle/debounce threshold? i really don't mind waiting 1s for linter to kick in if it means it's not continuously linting while i type (which combined w/ the hovering tooltip can really get in the way of seeing your context)

`linter:toggle` is per editor? that would also help with my "too many errors / panel hiding entire editor" gripe in #722

steelbrain:
We don't really keep a threshold in our codebase, we use the core's `onDidStopChanging` event. and we've found it to work great, if you have any performance issues with linters, we would recommend disabling `lintOnFly` so you only get results when you save a file.

nemoDreamer:
Yeah, I worked w/ that for a while, but it gets confusing fast when all the markings don't move w/ the line-changes, etc. But yes, it's definitely a valid workaround.

Could I change my core's `onDidStopChanging` threshold? Maybe there's a plug-in for that :wink:, will see before I start muckin'about.

steelbrain:
It's hard-coded I think.

mark-bradshaw:
[...]

PR #732, opened 2015-07-13
Fix bubble marker leak

steelbrain:
Just like the title says. We were leaking a marker, that would slow the text editor over time.
Ref: #708


AsaAyers:
Does the marker get cleared when the cursor moves, or when the next bubble gets placed?

steelbrain:
when the cursor moves, the previous one is removed and it's checked if a new one is appropriate.

AsaAyers:
I see where that is now. I just missed it.

PR #736, opened 2015-07-14
Isolate Editor-Linter

steelbrain:
No more passing `@linter` around.
The lint method in this PR is a no-op, I'll make it work in a follow-up where I port `linter-registry` from the ng branch.

:fire: Merge https://github.com/AtomLinter/Linter/pull/737 first :fire:

- [x] @park9140
  - [x] Make tests more readable


lib/editor-linter.coffee
```diff
@@ -1,102 +1,42 @@
...
-    @messages = new Map
-    @inProgress = false
-    @inProgressFly = false
-
-    if @editor is atom.workspace.getActiveTextEditor()
-      @linter.views.updateLineMessages(true)
-
+  constructor: (@editor) ->
+    throw new Error("Given editor isn't really an editor") unless @editor instanceof TextEditor
+    @status = true # Overall linting status
```
AsaAyers: Is this the same `@status` that changes when you use `Linter: toggle`? If so, do you mind if we call it `@enabled` instead? I think that makes the purpose more clear.

steelbrain: It's not actually being used anywhere, gonna remove it instead. I have used an alternative approach for `linter:toggle`

lib/editor-linter.coffee
```diff
@@ -1,102 +1,42 @@
...
-    )
-    @subscriptions.add(
-      @editor.onDidChangeCursorPosition ({oldBufferPosition, newBufferPosition}) =>
+    @subscriptions.add @editor.onDidDestroy =>
+      @emitter.emit 'did-destroy'
+
+    @subscriptions.add @editor.onDidSave => @emitter.emit('should-lint', false)
+    # The onDidStopChanging callbacks are invoked immediately, and we want to avoid it
+    setImmediate =>
+      @subscriptions.add @editor.onDidStopChanging => setImmediate => @emitter.emit('should-lint', true)
```
AsaAyers: I think this is just too complex for one line and all implicit braces. It takes a minute to figure out which of these are parameters and to which functions.

steelbrain: I find it quite straight forward one-liner, how would you suggest it to be written?

lib/editor-linter.coffee
```diff
@@ -1,102 +1,41 @@
...
-      @editor.onDidSave => @lint(false)
-    )
-    @subscriptions.add(
-      @editor.onDidChangeCursorPosition ({oldBufferPosition, newBufferPosition}) =>
+    @subscriptions.add @editor.onDidDestroy =>
+      @emitter.emit 'did-destroy'
+
+    @subscriptions.add @editor.onDidSave => @emitter.emit('should-lint', false)
+    # The onDidStopChanging callbacks are invoked immediately, and we want to avoid it
+    setImmediate =>
```
AsaAyers: What is this outer `setImmediate` for? Maybe I'm missing something, but it doesn't really look necessary. Also I think the comment above this line belongs on the `@subscriptions.add...` line

steelbrain: Everytime a text editor is created, triggering an onDidStopChanging event is one of the first things it does, and we don't want it for the first time. `setImmediate` works like a `process.nextTick` but without the 100 nested limit.
```coffee
setImmediate =>
  console.log("second")
console.log("first")
```
and You'll get `first` then `second` in your console.

lib/editor-linter.coffee
AsaAyers:
```CoffeeScript
emitShouldLint = =>
    # The onDidStopChanging callbacks are invoked immediately, and we want to avoid it
    setImmediate => @emitter.emit('should-lint', true)
    # Alternatively this will also work and doesn't require an intermediate function just for binding.
    # setImmediate @emitter.emit.bind(this, 'should-lint', true)


@subscriptions.add @editor.onDidStopChanging(emitShouldLint)
```

This separates registering the callback from the definition of it and makes it easier to see the two pieces.

steelbrain: We actually want to trigger the `should-lint` asynchronously as well, and You are doing it synchronously above.

lib/editor-linter.coffee
AsaAyers: but why do you have two `setImmediate`s? It seems like the one inside the `onDidStopChanging` solves that

steelbrain: The one inside it is to trigger the `should-lint` asynchronously, so the other handlers on that event don't get blocked because of it.

lib/editor-linter.coffee
AsaAyers: No, I've defined `emitShouldLint`, when called it will asynchronously trigger `should-lint`.

Now that we have an asynchronous emitter we can register that callback. The only difference between my code and yours is that I named the function instead of nesting 2 anonymous functions inside of a callback registration

steelbrain: We actually need that outer setImmediate as well, see my comment below. and can we talk in one point instead of two please, it's quite confusing :sweat_smile:

lib/editor-linter.coffee
AsaAyers: I see your comment and I still don't follow how it's needed.

`editorLinter = new EditorLinter(editor)` When this code executes the outer `setImmediate` is called. The rest of the `atom.workspace.observeTextEditors` callback completes, then we come back here and subscribe to `@editor.onDidStopChanging`. It calls the callback immediately, which triggers a 2nd setImmediate. At that point the rest of this function gets to complete (`@editor.onDidChangeCursorPosition`). After it completes then the callback executes that triggers `@emitter.emit('should-lint', true)`

So why can't adding these subscriptions happen during the construction of the class? the `should-lint` event is already asynchronous

lib/linter-plus.coffee
```diff
@@ -31,12 +31,17 @@ class Linter
...
-      editor.onDidDestroy =>
-        currentEditorLinter.destroy()
+      editorLinter = new EditorLinter(editor)
+      @editorLinters.set editor, editorLinter
+      @emitter.emit 'observe-editor-linters', editorLinter
+      editorLinter.onShouldUpdateBubble =>
+        @views.renderBubble()
+      editorLinter.onShouldUpdateLineMessages =>
+        @views.updateLineMessages(true)
+      editorLinter.onShouldLint =>
```
AsaAyers: Why isn't this just contained inside `editorLinter`? Why is `EditorLinter` even emitting an event instead of just calling `@lint` internally?

lib/editor-linter.coffee
steelbrain: I think you have a mis-understanding there, `@editor.onDidStopChanging` is only triggered immediately if we subscribe on construct, if we subscribe on `setImmediate` it no longer calls immediately.

lib/linter-plus.coffee
steelbrain: Because the linters are stored on the linter object and instead of changing or accessing your neighbours, I found this one to be easy to debug and specific. I mean, editor stuff and subscriptions are stored in editor-linter and linter stuff like addLinter, deleteLinter, lint and locking and stuff is managed by `linter-registry`, You can check it out in https://github.com/AtomLinter/Linter/pull/737

lib/editor-linter.coffee
AsaAyers: ok, I'm going to quit arguing, but something here is complex enough that I can't seem to understand it. That's definitely a code smell.

steelbrain: I think I am gonna add an explanation above it.

lib/editor-linter.coffee
```diff
@@ -1,102 +1,40 @@
...
-    )
+    @subscriptions.add @editor.onDidSave => @emitter.emit('should-lint', false)
+    @subscriptions.add @editor.onDidChangeCursorPosition ({oldBufferPosition, newBufferPosition}) =>
+      if newBufferPosition.row isnt oldBufferPosition.row
+        @emitter.emit('should-update-line-messages')
+      @emitter.emit('should-update-bubble')
+    # The onDidStopChanging callbacks are invoked immediately on creation, We are just
+    # gonna wait until a bit to get real events
+    setImmediate =>
+      @subscriptions.add @editor.onDidStopChanging => setImmediate => @emitter.emit('should-lint', true)
```
AsaAyers: I still don't think the outer `setImmediate` is necessary, but I'll give my :+1: on merging this because we aren't getting anywhere.

steelbrain: This PR actually can't be merged before we merge the ones it depends on :broken_heart:

AsaAyers: yup, which is why I didn't push the button.

I'm sick tonight so I won't be reviewing the others. Maybe someone else in @atom-community/linter-contributors can review #737 and #738?

lib/linter-plus.coffee
park9140: @steelbrain, can rephrase the last comment without using the word 'stuff'.

lib/linter-plus.coffee
```diff
@@ -110,7 +115,7 @@ class Linter
   deactivate: ->
     @subscriptions.dispose()
```
park9140: This is being disposed, implement the dispose interface.

lib/validate.coffee
```diff
@@ -0,0 +1,27 @@
+{Range} = require('atom')
+helpers = require('./helpers')
+
+module.exports = Validate =
```
park9140: Whitspace is your friend.  Please make this more readable.

spec/editor-linter-spec.coffee
```diff
@@ -0,0 +1,41 @@
+describe 'editor-linter', ->
+  EditorLinter = require('../lib/editor-linter')
+  beforeEach ->
+    waitsForPromise ->
+      atom.workspace.destroyActivePaneItem()
+      atom.workspace.open('test.txt')
+  describe '::constructor', ->
```
park9140: whitespace is your necessary for readablilty. separate describes and tests with empty lines. separate setup from expectation with whitespace.

spec/editor-linter-spec.coffee
```diff
@@ -0,0 +1,41 @@
...
+      timesTriggered = 0
+      editorLinter.onShouldLint ->
+        timesTriggered++
+      textEditor.save()
+      textEditor.save()
+      textEditor.save()
+      textEditor.save()
+      textEditor.save()
+      expect(timesTriggered).toBe(5)
+      editorLinter.deactivate()
```
park9140: editorLinter and textEditor are created for both tests. use beforeEach and afterEach to separate setup and teardown from the tests.

spec/validate-spec.coffee
```diff
@@ -0,0 +1,31 @@
+describe 'validate', ->
+  validate = require('../lib/validate')
+  describe '::linter', ->
```
park9140: whitespace is your necessary for readablilty. separate describes and tests with empty lines. separate setup from expectation with whitespace.

park9140:
:ship: After test cleanup.  Sorry it took so long to get to this.

steelbrain:
Will do, Thanks!

spec/editor-linter-spec.coffee
steelbrain: done

steelbrain:
Tests cleaned up, all green in local `apm test`

PR #737, opened 2015-07-15
Introduce Linter-Registry

steelbrain:
To be merged into #736

:fire: Merge https://github.com/AtomLinter/Linter/pull/738 first :fire:

##Reviewers
- [x] @park9140
  - [ ] Make tests more readable


steelbrain:
The actual code is less than a hundred lines, most of it is specs.

lib/linter-registry.coffee
```diff
@@ -0,0 +1,68 @@
...
+    return if onChange and not atom.config.get('linter.lintOnFly')
+    return unless editor is atom.workspace.getActiveTextEditor()
+    return unless editor.getPath()
+    return if @locks[lockKey].has(editorLinter)
+
+    @locks[lockKey].add(editorLinter)
+    scopes = editor.scopeDescriptorForBufferPosition(editor.getCursorBufferPosition()).scopes
+    scopes.push('*') # To allow global linters
+
+    # Confusing code ahead, proceed with caution :P
```
AsaAyers: Maybe it could be written in a way that's less confusing instead of just throwing a comment on it.

lib/linter-registry.coffee
```diff
@@ -0,0 +1,68 @@
...
+    return @linters.reduce((promise, linter) =>
+      if helpers.shouldTriggerLinter(linter, true, onChange, scopes)
+        return promise.then(-> linter.lint(editor)).then((vals) =>
+          if vals then @emitter.emit('did-update-messages', {linter, messages: vals, editor})
+        ).catch (e) -> helpers.error(e)
+      else return promise
+    , Promise.resolve()).then( =>
+      Promises = []
+      @linters.forEach (linter) =>
```
AsaAyers: if `@linters` had a `.map()` you can have that build up the `Promises` variable for you. Also, `Promise.all` doesn't require an array of promises. You can throw other values like `undefined` in and it'll work the same way

lib/linter-registry.coffee
```diff
@@ -0,0 +1,68 @@
...
+        return unless helpers.shouldTriggerLinter(linter, false, onChange, scopes)
+        try result = linter.lint(editor) catch e then return helpers.error(e)
```
AsaAyers: I recommend replacing this very awkward code with a form of what was here originally.

```CoffeeScript
        return new Promise((resolve) =>
          resolve(linter.lint(@editor, Helpers))
        ).then((results) => )
        .catch((e) -> helpers.error(e) # <- this line is new
```

1. the code above is easier to read
2. it doesn't have any of this checking to see if it's a promise and wrapping it nonsense.

steelbrain: done

lib/linter-registry.coffee
steelbrain: done

lib/linter-registry.coffee
steelbrain: done

park9140: Also comment is confusing as it just states danger, does not help understand code.   Remove it.

lib/linter-registry.coffee
```diff
@@ -0,0 +1,68 @@
...
+    return new Promise((resolve) =>
+      resolve(linter.lint(editor))
+    ).then((results) =>
+      if results then @emitter.emit('did-update-messages', {linter, messages: results, editor})
+    ).catch((e) -> helpers.error(e))
+
+  onDidUpdateMessages: (callback) ->
+    return @emitter.on('did-update-messages', callback)
+
+  deactivate: ->
```
park9140: As with other things. Use the disposable interface in order to be able to utilize the composite disposables at other levels in order to reduce boilerplate.

lib/linter-registry.coffee
```diff
@@ -0,0 +1,68 @@
...
+    @emitter.dispose()
+    @linters = []
```
park9140: If you are disposing this you should just use delete to free the reference rather than creating new memory.

lib/linter-registry.coffee
steelbrain: Acknowledged.

lib/linter-registry.coffee
steelbrain: I am not sure if I understand that :sweat_smile:

spec/linter-registry-spec.coffee
```diff
@@ -0,0 +1,169 @@
...
+  beforeEach ->
+    waitsForPromise ->
+      atom.workspace.destroyActivePaneItem()
+      atom.workspace.open('test.txt')
+  describe '::construct', ->
+    it 'expects no arguments', ->
+      linterRegistry = new LinterRegistry
+      expect(true).toBe(true)
+      linterRegistry.deactivate()
+  describe '::addLinter', ->
```
park9140: Whitespace is your friend and makes everything more readable.  Add some between describe blocks and test blocks

spec/linter-registry-spec.coffee
```diff
@@ -0,0 +1,169 @@
...
+    it 'adds error notification if linter is invalid', ->
+      linterRegistry = new LinterRegistry
```
park9140: separate setup into beforeEach where possible.

spec/linter-registry-spec.coffee
```diff
@@ -0,0 +1,169 @@
...
+      linter = {grammarScopes: ['*'], lintOnFly: false, modifiesBuffer: false, scope: 'file', lint: -> }
+      linterRegistry = new LinterRegistry
+      linterRegistry.addLinter(linter)
+      expect(linterRegistry.linters.length).toBe(1)
+      linterRegistry.deactivate()
+  describe '::hasLinter', ->
+    it 'returns true if present', ->
+      linter = {grammarScopes: ['*'], lintOnFly: false, modifiesBuffer: false, scope: 'file', lint: -> }
+      linterRegistry = new LinterRegistry
+      linterRegistry.addLinter(linter)
```
park9140: separate test specific setup from expectations with some whitespace.

spec/linter-registry-spec.coffee
```diff
@@ -0,0 +1,169 @@
...
+      expect(linterRegistry.hasLinter(linter)).toBe(true)
+      linterRegistry.deactivate()
+    it 'returns false if not', ->
+      linter = {grammarScopes: ['*'], lintOnFly: false, modifiesBuffer: false, scope: 'file', lint: -> }
+      linterRegistry = new LinterRegistry
+      expect(linterRegistry.hasLinter(linter)).toBe(false)
+      linterRegistry.deactivate()
+  describe '::deleteLinter', ->
+    it 'deletes the linter from registry', ->
+      linter = {grammarScopes: ['*'], lintOnFly: false, modifiesBuffer: false, scope: 'file', lint: -> }
```
park9140: this generic linter is used multiple times, you should create a named function to build it so that you don't repeat code and it's specific definition is better locked for people reading the tests.

spec/linter-registry-spec.coffee
```diff
@@ -0,0 +1,169 @@
...
+      expect(linterRegistry.hasLinter(linter)).toBe(true)
+      linterRegistry.deleteLinter(linter)
+      expect(linterRegistry.hasLinter(linter)).toBe(false)
+      linterRegistry.deactivate()
+  describe '::lint', ->
+    it "doesn't lint if textEditor isn't active one", ->
+      linterRegistry = new LinterRegistry
+      editorLinter = new EditorLinter(atom.workspace.getActiveTextEditor())
+      linter = {
+        grammarScopes: ['*']
```
park9140: case in point. this is formatted differently than the linter above but is configured the same. As a reader this confuses me and makes me waste time inspecting deeper for differences where none exist.

park9140:
:ship: after making tests more readable

steelbrain:
Acknowledged, Thanks!

lib/linter-registry.coffee
park9140: By using the disposable interface we can implement a more streamlined disposal handling externally to this class

```coffeescript
init ->
  @disposables = new CompositeDisposable();
  @linterRegistry = new LinterRegistry();
  @disposables.add(@linterRegistry);
  #event registrations and all things that are disposable all added to @disposables
dispose  ->
  @disposables.dispose(); #now this clear any set of things that we implement disposable for;
  #any class specific non disposable implementing logic
```

steelbrain: That is beautiful, any links about implementing that interface in JS?

lib/linter-registry.coffee
steelbrain: It actually frees the v8 memory so the garbage collector can garbage collect it easily. and linters is assigned to an empty array here, in case we dispose the editors and stuff before/after it and they trigger lint event.

lib/linter-registry.coffee
park9140: https://atom.io/docs/api/v1.0.2/Disposable#instance-dispose you can either use that class or just implement it's public properties.

steelbrain: Do you know of any good examples using this?

lib/linter-registry.coffee
park9140: What I meant was creating the new array just wastes memory unnecessarily if we are disposing this object.  If we have outstanding executions at this point then we have an issue with our event registrations such that we are not disposing them properly.

steelbrain: Agreed.

steelbrain: I tried it and it throws an error if there's a linting in progress, anyways `[]` takes an unnoticable amount of memory, I've seen a lot of tutorials and articles use `[].forEach` instead of `Array.prototype.forEach`

spec/linter-registry-spec.coffee
steelbrain: done

spec/linter-registry-spec.coffee
steelbrain: done

spec/linter-registry-spec.coffee
steelbrain: done

spec/linter-registry-spec.coffee
steelbrain: it actually has a different lint method, which modifies the value of a var

steelbrain:
All done

PR #745, opened 2015-07-16
Prevent linter message from wrapping

simurai:
Fixes:

screenshot_from_2015-07-16_08-09-11_360

### DIFF
--- a/lib/views/message.coffee
+++ b/lib/views/message.coffee
@@ -12,6 +12,7 @@ class Message extends HTMLElement
       return if message.filePath.indexOf(path) isnt 0 or displayFile isnt message.filePath # Avoid double replacing
       displayFile = message.filePath.substr( path.length + 1 ) # Remove the trailing slash as well
     el = document.createElement 'a'
+    el.classList.add 'linter-message-item'
     el.addEventListener 'click', ->
       Message.onClick message.filePath, message.range
     if message.range
@@ -22,6 +23,7 @@ class Message extends HTMLElement
 
   @renderRibbon: (message) ->
     el = document.createElement 'span'
+    el.classList.add 'linter-message-item'
     el.classList.add 'badge'
     el.classList.add 'badge-flexible'
     el.classList.add "linter-highlight"
@@ -31,6 +33,7 @@ class Message extends HTMLElement
 
   @renderMessage: (message, {cloneNode}) ->
     el = document.createElement 'span'
+    el.classList.add 'linter-message-item'
     if message.html
       if typeof message.html is 'string'
         el.innerHTML = message.html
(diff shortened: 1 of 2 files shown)


steelbrain:
LGTM. To fellow reviews: It does has a downside that the tooltip's tip is no more exactly on the spot but I say it's worth it.

simurai:
> It does has a downside that the tooltip's tip is no more exactly on the spot

Right. Actually even with wrapping, the tip can still be pushed away from the spot when there is little space.

simurai:
Alternatively, the whole lines could wrap too with:

```css
linter-message {
  flex-wrap: wrap;
}
```

tooltip

steelbrain:
Beautiful :heart:

simurai:
Now the alignment could still be improved. The 2nd line isn't aligned like the badge and 3rd line. I'll do that tomorrow.

simurai:
Ok, now each item wraps, but if there is even less space or the message is long, the text wraps too:

tooltip

I saw there is a `linter-multiline-message`. How does that show up? Haven't tested how that looks.

joefitzgerald:
Please take a look at ac+, which flips the direction of the tool tip when you approach the right.

park9140:
@simurai, as @joefitzgerald  suggests flipping the view would be better, additionally we should probably use the editor window boundry as the overflow so that we can use as much space as possible.

simurai:
@park9140 This PR fixes just the wrapping. Doesn't have to be final solution, just improvement over the current (broken) state.

Flipping would definitely be a good to have. Anyone wants to tackle this in another PR?

steelbrain:
@simurai should I merge this in?

simurai:
@steelbrain Sure, if you like it. As said, it just improves the current wrapping that looks kinda broken.

simurai:
btw. something I was thinking. Is it necessary for the inline message to show what line it is? Because if you click it, the only thing it does is go to the beginning, but it might not be that necessary.

steelbrain:
where do we show it then if not on the current line?

simurai:
It's still in the bottom panel, just not inline. Because inline, you already know what line and don't need a link.

screen shot 2015-07-23 at 10 27 10 am

steelbrain:
The inline bubble is there for the traces, that help the person get to the definition of the error object by just clicking a link.
screenshot from 2015-07-22 19-32-10

I use it everday and it boosts my productivity.

simurai:
> get to the definition of the error object by just clicking a link.

Ohh.. I see. Didn't know about that Trace thing. Ok, nevermind then.

nelson6e65:
> So removing the line info makes the tooltip a bit shorter.

Oh... I was thinking about it too. #776

PR #755, opened 2015-07-20
:fire: Comment 'alt' keybinding as a workaround to #648

samueloph:
It has passed more than 20 days by now of #648 report and It shouldn't be acceptable that the plugin break atom keybinding in such way for too long, I'm sending a PR commenting the keybinding on the plugin as a temporary workaround until we get a proper fix.


steelbrain:
How about you remove the line completely?

steelbrain:
Thanks

PR #757, opened 2015-07-20
Inline Messenger Service

mdgriffith:
Hi guys,

I recently put together an [inline messenger](https://atom.io/packages/inline-messenger) service for atom and I thought it might be a good fit for the linter project.

This pull request shows what linter would be like using the inline messenger.

Some small design differences from your current tooltips:
   * Mostly you can see what the messages look like here: [inline messenger](https://atom.io/packages/inline-messenger)
   * The bubble position can be set in the settings (below or to the right)
   * Message bubbles are set relative to the code they highlight instead of following the cursor. I believe this improves the experience.
   * The ability to make code suggestions that can be quickly accepted by the user.  I think this functionality would go great with linter. 
   * Multiline messages are highlighted along the gutter, similar to the git status gutter highlight.
   * Badges can be toggled on or off in the settings. 

Some points I know you guys might have concerns about
   * Introducing a new dependency
   * Your code used the same rendering html for panels as for inline messages

To address those, I'd like to give you guys commit access to the project, or talk about transferring ownership to atom-community, as long as I can join atom-community with commit access to this project.

I think the advantage here is in improving the atom ecosystem by having an inline messaging service that any package can use.  

Cheers and I really appreciate all the work you've put into linter.
-Matt

PR #767, opened 2015-07-23
Add Optional Accept Button to Linter Message

kaeluka:
Some linters (for instance haskell's hlint) suggest better code that
may replace offending expressions. It makes sense to have a button to
click on that removes the offending range and inserts the fix the
linter proposed.

http://imgur.com/ocTqVtk (see bottom right!)

This commit adds such a feature: if a message has an optional field
called `replacement` (of type string), the message window will offer
an "Accept!" option next to the source location. Clicking this option
replaces the marked range with the replacement string.

### DIFF
--- a/lib/views/message.coffee
+++ b/lib/views/message.coffee
@@ -5,6 +5,7 @@ class Message extends HTMLElement
     @appendChild Message.renderRibbon(@message)
     @appendChild Message.renderMessage(@message, @options)
     @appendChild Message.renderLink(@message, @options) if @message.filePath
+    @appendChild Message.renderAcceptReplacement(@message) if @message.replacement
 
   @renderLink: (message, {addPath}) ->
     displayFile = message.filePath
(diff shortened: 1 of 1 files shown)


steelbrain:
Well, we already have buffer modifying linters :sweat_smile: I am not sure if this is appropriate in the core, as we already support HTMLElements for `html` message property.
/cc @atom-community/linter-contributors

kaeluka:
Ha! No problem -- is the feature documented somewhere then?

steelbrain:
I just checked the docs, and it's not there, maybe they need to be updated.

kaeluka:
@steelbrain which linter is an example of one that modifies buffers?

steelbrain:
We don't really have any linter like that, but the API was requested by @joefitzgerald 
It doesn't really give it some helpers to modify the buffers, I just runs the linters that have `modifiesBuffer` in sequence, before the other linters. so they can perform certain "auto" corrections, without the user clicking something.

kaeluka:
My intention with this feature is NOT to support auto corrections. I'm looking at a lot of code here and only some linter suggestions make sense :)

The click is really essential in my case.

mdgriffith:
Fyi, there is support for suggestions in the [inline messenger](https://atom.io/packages/inline-messenger).  So, we may already be in line to support this once we move over to using it.

kaeluka:
That looks gorgeous! When is the move happening? Very nice that this functionality is being factored out.

mdgriffith:
Thanks!  The PR for this is here: https://github.com/atom-community/linter/pull/757

steelbrain:
@kaeluka mind closing this in favor of #757?

kaeluka:
Of course!

PR #770, opened 2015-07-25
Toggle Panel on double click

steelbrain:
Fixes https://github.com/atom-community/linter/issues/722


steelbrain:
Thinking of adding a test for it. so it doesn't break in the future.

park9140:
:Ship:

PR #778, opened 2015-07-29
feat: use current file to detect syntax

### LINKED ISSUE #777: Markdown get linted
https://github.com/AtomLinter/linter-jshint/issues/136
https://github.com/AtomLinter/linter-jscs/issues/62

And I can reproduce it too with `linter-eslint`

Example code:

```markdown

# Some code example with syntax will be linted

```javascript
function () {}
```

The problem is that the whole file is getting linted so it throws a lot errors.

iam4x:
fixes #777

### DIFF
--- a/lib/linter-registry.coffee
+++ b/lib/linter-registry.coffee
@@ -35,8 +35,9 @@ class LinterRegistry
     return if @locks[lockKey].has(editorLinter)
 
     @locks[lockKey].add(editorLinter)
-    scopes = editor.scopeDescriptorForBufferPosition(editor.getCursorBufferPosition()).scopes
-    scopes.push('*') # To allow global linters
+
+    # Add scope of currentFile and allow global linters
+    scopes = [editor.getGrammar().scopeName, '*']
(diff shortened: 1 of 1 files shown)


steelbrain:
A test please.

steelbrain:
@joefitzgerald I am not entirely sure if this would be correct as it doesn't report `source.php` for a php file. Any ideas, also how did you do it in AC+?

iam4x:
@steelbrain Maybe we can have a special case for handling `.md` files? Instead of breaking actual way to find file syntax.

steelbrain:
@iam4x but that would be monkey patching, I would prefer to fix this issue once and for all :)

iam4x:
Let's wait for advices :)

steelbrain:
Beautiful :)

steelbrain:
No activity in a long long time

steelbrain:
Also fixed in Atom by https://github.com/atom/language-gfm/pull/137

PR #781, opened 2015-07-31
Pull the plug

steelbrain:
Just like the title says
Merging this into `next` branch instead of master


steelbrain:
Stuff's green, merging in.

k2b6s9j:
Shouldn't we at least throw a warning when someone attempts to use a legacy linter?

steelbrain:
Well.... We should. I have a few things to discuss, waiting for you on Slack.

PR #782, opened 2015-07-31
Show a warning for depreciated ones

steelbrain:
They won't work, just a warning.
/cc @k2b6s9j

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/782)

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -42,6 +42,14 @@ module.exports =
   activate: (state) ->
     LinterPlus = require('./linter-plus.coffee')
     @instance = new LinterPlus state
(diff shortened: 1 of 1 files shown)


k2b6s9j:
Is deprecation cop the best place to do this warning? I mean, aren't deprecations meant only as warning for code that is about to change, not code which has already changed?

k2b6s9j:
Although, now that I think about it I've seen iOS throw Deprecation Warnings for methods that no longer work...

k2b6s9j:
Yeah. Looks good to me. This will start a new line of issues being created telling provider maintainer that their code will no longer run.

PR #783, opened 2015-07-31
Cleanup filtering

steelbrain:
We are no longer filtering messages on get, filtering them on set instead.

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/783)


steelbrain:
I might add a spec for it as well

steelbrain:
Travis's green. Merging in

PR #789, opened 2015-08-02
Allow stringish errors

steelbrain:
Atom-Linter module at the moment throws stringish errors, I am gonna fix this in that module as well. But thought we should be handling this scenario


steelbrain:
All green via `apm test`, don't wanna wait for Travis.

PR #820, opened 2015-08-10
Write message-registry in ES6

steelbrain:
It fixes a lot of typos, it doesn't self-invoke in the `updatePublic` method now, we invoke a separate updater, that's easily controllable.

It also fixes some scenarios where some removed messages would still be shown as markers.

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/820)


steelbrain:
Everything's green, merging in.

PR #827, opened 2015-08-12
Efficient Panel and Bugfixes

steelbrain:
Fixes a weird bug with Linter registry where the old messages in the messages array would be new objects, which led to some cases where markers weren't destroyed. Also fixes a bug where if we had two messages of the same JSON representation, they would be treated as same, meaning if we had two markers for them, only one would be destroyed. Other would become a zombie.

Also introduces super-efficient panel updates.
Also fixes a spec.
Also fixes an invalid token error when adding a class for ribbon

ToDo:
 - [x] Add specs for new classes
 - [x] Make expanding of multi-line-messages work

Fixes https://github.com/atom-community/linter/issues/822
Partially Fixes https://github.com/atom-community/linter/issues/708

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/827)


steelbrain:
This is done and stable. Gonna fix the markers bug now.

Issue #575, opened 2015-06-23
Status Bar Settings


jonathandelgado:
Rewrite looks really interesting, great job guys!

Any way to disable the "Current File" / "Project" tabs? It's a little bit too much clutter for me. Perhaps it should be a setting rather than a fast toggle?

Another quick question, would you consider adding a setting to put the "No Errors" message on the right side of the sidebar again? Most other plugins use the right side, so it just looks odd to me.

Thanks for your time!

steelbrain:
:+1: for having it on the right sidebar.
:-1: for being able to disable the tabs. Can you please provide more explanation on why would you want it? It doesn't take much space and will be very helpful to you in case you use FlowJS.

jonathandelgado:
@steelbrain Thanks for the reply!

Glad you agree with my second comment. For the tabs, it just seems out of place, i've never seen buttons on the sidebar. I guess i'm just really overly sensitive about my text editor. :) No worries though, I'll just make a plugin that deletes the node on load. Just wanted to make the issue in case you agreed.

Thanks again!

steelbrain:
@jonathandelgado If they seem out of place to you, Can you suggest an alternative layout for it?

Also, just added a PR to show the StatusBar on the right.

arbourd:
:+1: for being able to disable the errors-list tab in settings (or at least toggle it to hide until I want to see it). I prefer the inline errors with a message at the bottom that shows how many.

Rubocop can be a little overzealous at times.

jonathandelgado:
@steelbrain Wow, thanks so much. I really appreciate it.

The only thing I can think of for an alternative layout would be simply reverting to a toggle on the command pallet, which I am sure would hinder the use-case you are going for. I guess a hide setting would be the most useful?

mnquintana:
> Can you suggest an alternative layout for it?

This isn't an alternative layout per se, but I think it would feel a lot less cluttered if the tabs were icons with tooltips explaining what they were, rather than full text.

steelbrain:
@mnquintana @simurai had actually suggested it https://github.com/AtomLinter/linter-plus/pull/76#issuecomment-113770672

pk-nb:
I too prefer just gutters and a count of errors (file only) in the bottom bar. Something you can add to your stylesheet in the meantime:

```less
#linter-panel,
.linter-tab,
.status-error {
  display: none;
}
```

Unfortunately the error count seems stuck on project, so I hide that too.

[...]

PR #837, opened 2015-08-16
Reverted ability to select status-bar side

### LINKED ISSUE #720: Changing position of status icon on bottom bar has no effect
Linter 1.2.3, Atom 1.0.2, OS X 10.10.4

I tried moving the status icon to the right side of the bottom bar, but it stayed on the left side. My `config.cson` has the following:

``` cson
"*":
  linter:
    statusIconPosition: "Right"
```

This option was added in #642. However it looks like it is now [hard coded to add the icon to the left](https://github.com/AtomLinter/Linter/blob/56521d704bf5fb09092bf3cf6edebff2746a562c/lib/linter-views.coffee#L120):

``` coffee
class LinterViews
  attachBottom: (statusBar) ->
    @bottomBar = statusBar.addLeftTile
```

I think #664 introduced the regression. Furthermore, I don't see how [the tests for this feature](https://github.com/AtomLinter/Linter/blob/56521d704bf5fb09092bf3cf6edebff2746a562c/spec/config/status-icon-position-spec.coffee#L44-L49) can possibly be passing:

``` coffee
describe "The Status Icon Configuration Option", ->
  describe "when set to 'Right'", ->
    it "is on the right side of the Status Bar.", ->
      jasmine.attachToDOM(workspaceElement)
      [linterHighlight] = statusBar.getRightTiles().map (tile) -> tile.getItem()
      expect(linterHighlight).toBeDefined
      [linterHighlight] = statusBar.getRightTiles().map (tile) -> tile.getItem()
      expect(linterHighlight).not.toBeDefined
```

Note that both checks use `getRgihtTiles`.

jonathandelgado:
Just reverted the `statusIconPosition` setting from e36281e3d4e8cfdae29c71ac790ebdfb1ae58030 and updated the attach function since there is only one container now. Additionally, the dynamic priority assures the tile will snap to the corresponding side.

Fixes #720 and resolves the regression in #575, #576, #630

Thanks for your time!

### DIFF
--- a/lib/linter-views.coffee
+++ b/lib/linter-views.coffee
@@ -109,10 +109,12 @@ class LinterViews
       ) if @underlineIssues
 
   attachBottom: (statusBar) ->
-    statusIconPosition = atom.config.get('linter.statusIconPosition')
-    @bottomBar = statusBar["add#{statusIconPosition}Tile"]
-      item: @bottomContainer,
-      priority: if statusIconPosition == 'Left' then -100 else 100
+    @subscriptions.add atom.config.observe('linter.statusIconPosition', (statusIconPosition) =>
+      @bottomBar?.destroy()
+      @bottomBar = statusBar["add#{statusIconPosition}Tile"]
+        item: @bottomContainer,
+        priority: if statusIconPosition == 'Left' then -100 else 100
+    )
 
   removeMarkers: (messages = @messages) ->
     messages.forEach((message) =>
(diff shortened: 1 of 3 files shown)


steelbrain:
How about you add a note to the description that this setting requires a restart to update. Also can you add a spec for this?

jonathandelgado:
To your previous comment, I actually just did an observe on the config to eliminate the restart requirement., that should be a lot smoother.

Additionally, added the spec as requested. Sorry for it being a little rough around the edges, hard to write clean code when querying the DOM.

steelbrain:
LGTM Thanks

jonathandelgado:
@steelbrain Thanks!

PR #843, opened 2015-08-17
Fix bottom container count

steelbrain:
refresh it on pane switch
Fixes https://github.com/atom-community/linter/issues/839

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/843)


steelbrain:
What's left is a spec

PR #851, opened 2015-08-18
Fix the next-previous implementation to account for cursor position

### LINKED ISSUE #833: `Linter: Next Error` jumps to seemingly random error
"next" should be relative to the cursor, but I think it's jumping based on some internal ordering or internal caching.

AsaAyers:
Fixes #833

### DIFF
--- a/lib/commands.coffee
+++ b/lib/commands.coffee
@@ -12,9 +12,6 @@ class Commands
       'linter:expand-multiline-messages': => @expandMultilineMessages()
       'linter:lint': => @lint()
 
-    # Default values
-    @index = null
-
   togglePanel: ->
     atom.config.set('linter.showErrorPanel', !atom.config.get('linter.showErrorPanel'))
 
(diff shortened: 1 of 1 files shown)


lib/commands.coffee
```diff
@@ -60,30 +57,49 @@ class Commands
...
-    else
-      @index = 0
-    message = @getMessage(@index)
+    filePath = atom.workspace.getActiveTextEditor().getPath()
+    {row, column} = atom.workspace.getActiveTextEditor().getCursorBufferPosition()
+    thisFile = @linter.views.messages.filter((m) -> filePath is m.filePath )
+
+    message = do ->
+      for m in thisFile
+        {start} = m.range
```
steelbrain: `continue unless m.range`

lib/commands.coffee
```diff
@@ -60,30 +57,49 @@ class Commands
...
+        if start.row > row
+          return m
+        if start.row is row and start.column > column
+          console.log('col', start.column, column)
```
steelbrain: You can just drop this line

lib/commands.coffee
AsaAyers: When do messages not have a range? Isn't it required?

steelbrain: Nope, not required.

lib/commands.coffee
```diff
@@ -60,30 +57,48 @@ class Commands
...
     atom.workspace.open(message.filePath).then ->
       atom.workspace.getActiveTextEditor().setCursorBufferPosition(message.range.start)
 
   previousError: ->
-    if @index?
-      @index--
-    else
-      @index = 0
-    message = @getMessage(@index)
+    filePath = atom.workspace.getActiveTextEditor().getPath()
```
steelbrain: `return unless filePath`

lib/commands.coffee
```diff
@@ -60,30 +57,48 @@ class Commands
...
-    # That way the index can be ++ an -- without caring about the array bounds.
-    messages[index %% messages.length]
-
   nextError: ->
-    if @index?
-      @index++
-    else
-      @index = 0
-    message = @getMessage(@index)
+    filePath = atom.workspace.getActiveTextEditor().getPath()
```
steelbrain: `return unless filePath`

steelbrain:
by the way, You don't have to filter messages like that, messages are already classified, You can just do `linter.views.messages.filter((m) => m.currentFile)`

AsaAyers:
ok, that's better than the early return because it maintains the ability to jump to another file that does have an error.

lib/commands.coffee
```diff
@@ -60,30 +57,46 @@ class Commands
...
-    # That way the index can be ++ an -- without caring about the array bounds.
-    messages[index %% messages.length]
-
   nextError: ->
-    if @index?
-      @index++
-    else
-      @index = 0
-    message = @getMessage(@index)
+    {row, column} = atom.workspace.getActiveTextEditor().getCursorBufferPosition()
```
steelbrain: what if `activeTextEditor` is null? It'll throw an `undefined is not a function error`

steelbrain:
Looks good, now all we need is some specs.

wmayner:
+1

AsaAyers:
Specs would be nice, but I'm not really sure how to test this. Can we get this merged without tests?

steelbrain:
I've modified the views and editor-linter classes a bit since this was opened. and You no longer have to filter all of the messages. If you do [`EditorLinter::getMessages()`](https://github.com/atom-community/linter/blob/527d2b3cadf4d9e209cd5d28505560aa3a32d3e7/lib/editor-linter.js#L122-L124) it should be significantly faster than iterating all of the messages. Also the object returned by it is an ES6 Map instead of an array, so you'll have to use the ES6 `for...of` (Coffee's `for...of` would fail) or the `Map.prototype.forEach` function.

Issue #871, opened 2015-08-30
Stale lint results may appear from use of multiple panes because MessageRegistry keys results by TextEditor instead of TextBuffer


asutherland:
The scenario is this:
* Be using 2 or more panes
* Have a file, let's say foo.js open in both panes, we'll call them Foo1 and Foo2.
* Edit Foo2 causing lint errors.
* Switch to Foo1, correcting the lint errors.  The errors from Foo2 will still be displayed while editing Foo1, using the same offsets, which is very confusing.
* Switching back to Foo2 and forcing a re-lint will cause the errors to no longer be displayed in Foo2 or Foo1.

I believe the problem is that:
* MessageRegistry persists messages on a per-TextEditor basis in `editorMessages`.
* Each pane gets its own TextEditor instance of the file.
* BottomPanel receives a list of all reported messages, and message visibility is decided based on `message.currentFile` which is a test of the current editor's path rather than editor identity.

It seems like the ideal solution is to key results based on the TextBuffer backing the TextEditor (which is accessible via TextEditor::getBuffer).  Alternately, results could be keyed on file path.  I think the primary risk in any keying is avoiding getting confused by renames, deletes, and buffers closing.

Since TextBuffers are inherently not aware of cursor positions, EditorLinter would still want to exist, so the simplest solution would likely be to just have MessageRegistry extract and use the buffers as it hears about them.  And for simplicitly deleteEditorMessages would err on the side of removing the buffer-keyed-messages, although something fancier involving some type of refcounting or direct observation of the TextBuffer life-cycle could be implemented.

I can provide a pull request, but since there are some architectural decisions involved and nuances I'm likely not aware of, I figured I'd file an issue first before trying to implement the fix.  (I, of course, will not take offense if the problem is magically fixed without me doing any work! :wink: )

PS: Thanks for this great package!

steelbrain:
I would very much welcome a PR for this. It's always good to get more and more people involved into this package.
About the architectural decisions, the purpose of EditorLinter is to store messages per editor and make sure that when linting is being done in one editorLinter it shouldn't re-trigger it unless previous invocation is done. I am not yet sure if we should use textbuffers instead of text editors there, but in message registry, sure, we should be using text buffers there.

PR #873, opened 2015-09-01
Messages per buffer fix

asutherland:
I think this implements what we discussed in #871.  Note that the tests are currently a separate commit since it lets me (and you if you like) easily verify that the current implementation fails the test and that with the fix everything is happy.  I would squash them after addressing any concerns you have prior to requesting final landing.  (Or you can squash them or do whatever you like.)

Please do provide idiomatic guidance on the spec file if it's not to your liking.  I don't do a lot of BDD testing and I worry that I may have done too much prep logic.  (I could see adding more it() cases for the multiple pane case like covering editor deletion, which is why I factored it to be more reusable.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 * Fix baseline alignment of text on the bottom bar
 * Fix a bug where linter:togglePanel won't work
 * Stricter message validation, helps catching provider bugs
+* Messages are now tracked on a per-buffer basis rather than per-editor,
+  improving the multi-pane experience by eliminating the potential for stale
+  results.
 
 ## 1.4.3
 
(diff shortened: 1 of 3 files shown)


steelbrain:
This looks good to me. We never squash commits by the way. and You do have a little more "logic" in the prep part but it's fine to me as their purpose is to just test that stuff works.

A changelog entry would be nice before I merge it in

asutherland:
I have pushed a changelog entry.  Also, glad to hear about the (lack of) squashing.

steelbrain:
Ouch, I forgot to update the changelog about the recent release, would you mind merging master in?

asutherland:
I did the non-history-rewriting thing.  Let me know if you do want some rewriting.

steelbrain:
It's all fine, just waiting on CI

steelbrain:
Thanks for your awesome contribution :clap:

asutherland:
And thank you for the very pleasurable contribution experience!

PR #884, opened 2015-09-15
Add `linter.displayLinterInfo` pref

devoncarew:
There are a few commits here, but the main work is to add a new `displayLinterInfo` preference. This lets users disable the linter contribution to the status bar. They can already do something similar for the error panel and the inline editor contributions. Some details:

- encode the order into the prefs
- add a new `displayLinterInfo` pref, and add code to observe its value
- normalize the titles and descriptions a bit
- fix some formatting and indentation nits

before:
screen shot 2015-09-15 at 1 29 03 pm

after:
screen shot 2015-09-15 at 1 42 51 pm


lib/main.coffee
```diff
@@ -7,47 +7,68 @@ module.exports =
...
+      title: 'Show Inline Tooltips'
+      description: 'Show inline tooltips for errors'
+      type: 'boolean'
+      default: true
+      order: 3
+    underlineIssues:
+      title: 'Underline Issues'
+      type: 'boolean'
+      default: true
+      order: 3
```
Arcanemagus: This has the same order as the item above it. (Which throws all subsequent items off as well.)

Arcanemagus: Actually, it looks like that works okay, as it would then just sort them alphabetically: https://github.com/atom/settings-view/pull/273
If you are going to the trouble of specifying an order though, why not specify all?

devoncarew: I'm using the order as more of a category bucket here, esp. where the alphabetical order works within the categories. Explicitly ordering everything could be a bit of a maintenance issue, but happy to do it either way.

steelbrain:
The implementation looks nice. Now all we need is a few specs.

devoncarew:
Cool, will add a few based on the existing ones.

devoncarew:
Added 3 tests -

spec/config/display-error-status-spec.coffee
```diff
@@ -0,0 +1,11 @@
+describe "The Status Bar Visibility Configuration Option", ->
```
steelbrain: You can just remove this spec

devoncarew: fair enough -

steelbrain:
LGTM. Thanks for your awesome contribution :clap:

PR #887, opened 2015-09-18
Fixed gutter corruption on Atom 1.0.13 - And looks better

JohnMurga:
Under recent versions of Atom setting margins or font sizes in the butter will cause alignment issues
This fixes that ... The dot was kinda large anyway, so I think it looks better now

### DIFF
--- a/styles/linter-plus.less
+++ b/styles/linter-plus.less
@@ -138,7 +138,6 @@ atom-text-editor::shadow .linter-highlight .icon-right{
   visibility: visible;
   &::before{
     content: @primitive-dot;
-    font-size: 1.1em;
   }
 }
 // Styling Error Types


steelbrain:
Thanks. :clap:

steelbrain:
Umm, reconsidering it, the dot now looks worse :(

steelbrain:
Can you please post your screenshots of before and after?

JohnMurga:
The first one is how it is now ...
Second one is before the change.
after_before

steelbrain:
I am not sure about you but on my screen the dot after this change looks super-tiny.

JohnMurga:
As you can see in the screenshot, it is smaller but looks better than the alternative ...
I think using "alert" instead of "primitive-dot" also looks OK ... But I might be weird :-)

steelbrain:
How about we replace the dot with a square?

JohnMurga:
Great :+1: :-)

steelbrain:
I remember having a square before but then some people said that it was too "loud". I think I'll look in an alternative strategy.

JohnMurga:
It'll be smaller now, so it might not be too loud anymore

steelbrain:
See https://github.com/atom-community/linter/issues/889

JohnMurga:
Nice feature ...
But I actually like the gutter icon :+1:

steelbrain:
In that case I guess you can always override styles in your local styles :)

JohnMurga:
Of course ;-)
I'll do that now
But other users on newer versions of Atom will have the line number issue from the screenshot

steelbrain:
I am not able to reproduce this locally, what's your screen resolution?

JohnMurga:
On that machine it was 1920x1280
With Atom 1.0.13 released 5 hours ago
There was gutter issues with a couple of of extensions that were fine on Atom 1.0.11
I removed the others, but kinda need this one :-)

steelbrain:
I'll try that release now.

steelbrain:
Ah. I can repro this :(

JohnMurga:
I guess my fix makes sense now ;-)

steelbrain:
Mind PRing again?

JohnMurga:
It would be a pleasure :-)
But I am away so it would be in about 6 hours.
I believe you can also unrevert the pull request.
Or manually remove the one offending line.

steelbrain:
Relanded in #890

PR #891, opened 2015-09-18
An alternative approach to coloring the gutter

JohnMurga:
Probably a dumb idea ... But I like how it looks ... 
untitled
It also does not mess with the folding, which the current icons do

### DIFF
--- a/styles/linter-plus.less
+++ b/styles/linter-plus.less
@@ -16,6 +16,8 @@ linter-message {
 
 .linter-message-item {
   margin: @spacing;
+  height: 1.8em;
+  white-space: nowrap;
 }
 
 linter-multiline-message {
@@ -134,6 +136,12 @@ linter-bottom-tab {
   }
 }
 
+atom-text-editor::shadow .linter-highlight .icon-right{
+  visibility: visible;
+  &::before{
+    content: @primitive-dot;
+  }
+}
 // Styling Error Types
 atom-text-editor::shadow .linter-highlight, .linter-highlight{
   &.badge {
@@ -146,7 +154,9 @@ atom-text-editor::shadow .linter-highlight, .linter-highlight{
       background-color: @background-color-error;
       color: white;
     }
-    color: @background-color-error;
+    .icon-right::before{
+      color: @background-color-error;
+    }
     .region {
       border-bottom: 1px dashed @background-color-error;
     }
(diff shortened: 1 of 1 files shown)


steelbrain:
I personally don't like the approach as the errors are not much visible :(

JohnMurga:
I did also try changing the background and bold-ness ...
But subtle seemed best

JohnMurga:
Could underline and strike-through also I guess

steelbrain:
I was considering making gutter optional and introducing an option to show inline bubbles on hover. That would simplify a lot of things.

Arcanemagus:
@JohnMurga how do the current icons mess with folding btw? In the theme I'm using at least it will simply condense the coloring down into a single (worst case) color:
image

JohnMurga:
@Arcanemagus I think was getting the "dot" icon, and it overrode the fold icon ...
Although coloring the fold icon would annoy me too :-)

JohnMurga:
Which is why I have : ```
atom-text-editor::shadow .gutter .foldable .icon-right {
    color: rgb(255,255, 255);
}
```
In my styles file

JohnMurga:
Ooops

steelbrain:
@Arcanemagus @SpainTrain @k2b6s9j Guys, what do you think about this gutter marker style? https://atom.io/packages/atom-wallaby

SpainTrain:
wallaby markers are pretty nice.  any idea how it looks when a line has a linter and wallaby marker?  Any update to the linter marker would ideally coalesce multiple markers nicely ^_^

k2b6s9j:
I don't think my opinion on this will be particularly helpful... I haven't had any problems with the gutters so far, and for the most part I really only make use of `minimap-linter`. Wish I could be more helpful.

Arcanemagus:
Beyond changing the current circle (which I have _no_ issues with) to a square, what does that change @steelbrain?

steelbrain:
@Arcanemagus at the moment we're styling the line number gutter and adding our styles to it's .icon-right::before, if some other package decides to use it we'll be in a conflict. Atom has a new API where you can add new gutter row completely independent, where we won't have any issues at all with the line-number one and where we can style the way we want it.

Arcanemagus:
@steelbrain that sounds awesome, and I see absolutely no reason not to move to it. That's beyond the scope of this PR, and should be filed as an issue though.

I was simply commenting on the visual style, as that is what this PR was about :wink:.

Issue #859, opened 2015-08-20
error text not aligned


jaypanares:
Noticing the following alignment issues.  On Windows 10 x64, Atom 1.0.7 and using linter-eslint package.

![capture](https://cloud.githubusercontent.com/assets/7044596/9379737/ec08930a-475d-11e5-840c-7ad0db2ffc60.JPG)

Alignment is still off even if I use default font settings.

steelbrain:
A PR would be appreciated.

steelbrain:
Closing as fixed

PR #892, opened 2015-09-18
Fix for #859

JohnMurga:
A two line fix to #859


steelbrain:
Screenshots of before and after would be appreciated

JohnMurga:
untitled

steelbrain:
Nice. A changelog entry would be appreciated, write it under `Upcoming` title

JohnMurga:
Done ... Is that what you meant ?

steelbrain:
LGTM Thanks

PR #899, opened 2015-09-22
Handle markers in editor-linter

steelbrain:
Features:
  - Markers will be available in different editors at the same time
  - Everytime a tab is switched, we remove all the markers from previous tab and re-add all the markers in new tab, keeping them in both would mean no unnecessary operations.
  - more that I can't think of right now

This does come at a speed regression. I'll be fixed it afterwards.

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/899)

### DIFF
--- a/lib/editor-linter.js
+++ b/lib/editor-linter.js
@@ -13,6 +13,9 @@ export default class EditorLinter {
     this.markers = new WeakMap()
     this.subscriptions = new CompositeDisposable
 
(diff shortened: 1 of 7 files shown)


steelbrain:
Okay, so the performance regression has been fixed. Now all we need to do is write some specs.

steelbrain:
Travis is green, specs are in, no performance wise effect. Merging it in.

PR #901, opened 2015-09-24
Add a custom gutter

steelbrain:
The API added in #899 is gonna be the base and we're gonna add own our gutter along with a few configurations for them. This will free us from a lot of styling limitations and won't cost much.

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/901)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Upcoming
 
+* Added `onDidMessage{Add, Remove, Change}` listeners on EditorLinter for self-api consumers
 * Linter now supports decorating multiple panes at the same time. Decorations are no longer removed and re-added on tab changed, only added to the new tab. Which could improve the tab switch performance with large errors.
 * Multiline messages render correctly by allowing overflow and using flexbox to enable single line output of the location.
 * Remove location from bubble information.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Removed EditorLinter::destroy in favor of EditorLinter::dispose
 * `.icon-right::before` css selector has been replaced by `.linter-gutter` of the new linter gutter element.
-* Added `onDidMessage{Add, Remove, Change}` listeners on EditorLinter for self-api consumers
+* Added `onDidMessage{Add, Remove, Delete}` listeners on EditorLinter for self-api consumers
 * Added EditorLinter::getMessages API for getting messages specific to that editor linter.
 * Linter now supports decorating multiple panes at the same time. Decorations are no longer removed and re-added on tab changed, only added to the new tab. Which could improve the tab switch performance with large errors.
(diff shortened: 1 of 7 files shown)


steelbrain:
I've added the required APIs and they work. This will be a css-selector breaking changes so we should ping the provide authors to update them before publishing. Also a few specs are to be wrote/fixed.
I'm waiting on @simurai to work his magic and make the UI beautiful again

simurai:
Is there a way to remove the inline `height: 24px`? It's not really needed, but I couldn't find from where it gets added.

screen shot 2015-09-25 at 2 06 24 pm

steelbrain:
@simurai that's actually added by atom, we can override it with a `!important`, that's the only way around it.

simurai:
Hmm.. strange, because unlike the parent, it doesn't get updated when font size is changed.. 

but yeah... `!important` saves the day.. again! :smile_cat:

simurai:
This PR is **only** about the custom gutter, right? I think it's good:

screen shot 2015-09-25 at 2 21 15 pm

The inline message for missing files could probably improved a bit, but I guess that should be a separate PR.

screen shot 2015-09-25 at 2 22 16 pm

steelbrain:
Well. That looks nice :+1:  Now all we have to do is write some specs for the new APIs that've been added and it'll be merged in. I am sure a lot of people are gonna like the ability to change the gutter highlight position.

simurai:
> I am sure a lot of people are gonna like the ability to change the gutter highlight position.

Do you think people/linters will want custom gutter highlights? I mean other than the "red dot".

This could be an easter egg:

screen shot 2015-09-25 at 2 37 12 pm

```less
atom-text-editor::shadow .linter-gutter {
  width: 1em !important;
  height: 1em !important;
  background: none !important;
  line-height: 1;
  &:before {
    content: "👹";
    font-size: .8em;
  }
}
```

steelbrain:
That's an awesome idea, but we should have an option for that's off by default :stuck_out_tongue:

simurai:
> off by default

Sure! Or people can just add to `styles.less`.

steelbrain:
Good Good

simurai:
Something to consider: Clicking the dot to open the inline message? Currently you have to click on the underline.

steelbrain:
That's a nice feature request, but how will we handle multiple messages on one line with that?

simurai:
You can cycle through them. For every click it jumps to the next and then starts at the first again.

Just a random thought. Doesn't need to be part of this PR.

steelbrain:
All tests seem to pass locally no idea where travis is failing, anyways, Imma add those easter eggs :stuck_out_tongue:

steelbrain:
On another thought, I just want to land this in. so I'll add the easter eggs in another PR or wait for the community to contribute and just work on this one's specs and merge it in.

steelbrain:
Travis is green. Specs are in. works as intended. Merging in.

PR #902, opened 2015-09-24
Fix linter message styles

### LINKED ISSUE #893: Multiline messages don't appear to work for me.
What am I doing wrong? I've already obliterated my Atom configuration and restarted.

https://www.youtube.com/watch?v=LAgrLTHZT_E

* The disclosure buttons don't appear to work.
* When I use *Expand Multiline Messages* from the command palette and press *Return*, the messages appear briefly and disappear in a flash.
** When I click *Expand Multiline Messages*, it works. 

Here's my config.cson. 

```cson
"*":
  welcome:
    showOnStartup: false
  core:
    disabledPackages: [
      "vim-mode"
      "vim-surround"
      "file-icons"
    ]
  editor:
    invisibles: {}
    fontSize: 11
  "one-dark-ui": {}
  "linter-rust":
    jobsNumber: 8
  linter: {}
```

blakeembrey:
Fixes #893. Related to #892 and #859.

image

On a related note, I did not find any output of the `multiline` class so this is purely the CSS fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Upcoming
 
 * Linter now supports decorating multiple panes at the same time. Decorations are no longer removed and re-added on tab changed, only added to the new tab. Which could improve the tab switch performance with large errors.
+* Multiline messages render correctly by allowing overflow and using flexbox to enable single line output of the location.
 
 ## 1.6.0
 
--- a/lib/ui/message-element.js
+++ b/lib/ui/message-element.js
@@ -37,7 +37,7 @@ export class Message extends HTMLElement {
     const pathEl = document.createElement('span')
     let displayFile = message.filePath
 
-    el.className = 'linter-message-item'
+    el.className = 'linter-message-item linter-message-link'
 
     for (let path of atom.project.getPaths())
       if (displayFile.indexOf(path) === 0) {
--- a/styles/linter-plus.less
+++ b/styles/linter-plus.less
@@ -5,19 +5,21 @@
 
 linter-message {
   display: flex;
-  align-items: flex-end;
-  flex-wrap: wrap;
+  align-items: flex-start;
   font-size: @font-size;
   line-height: 1.2;
-  & > .badge {
-    align-self: flex-start;
-  }
 }
 
 .linter-message-item {
   margin: @spacing;
-  height: 1.8em;
-  white-space: nowrap;
+}
+
+.linter-highlight {
+  flex: 0 0 auto;
+}
+
+.linter-message-link {
+  flex: 1 0 auto;
 }
 
 linter-multiline-message {


steelbrain:
before and after images for the comparison would be nice to have. along with a changelog entry under the title `Upcoming`

blakeembrey:
On a related note, it might be a good idea to have a max width for the location link. I'll add the extra information now.

blakeembrey:
@steelbrain Added, and here's the before:

image

steelbrain:
LGTM Thanks :heart:

blakeembrey:
Awesome :+1: Just a side note too, but you can change the `link` class to `flex: 0 0 auto` to have it align to the right like:

image

Might be a little cleaner in some cases to have it align nicely.

blakeembrey:
@steelbrain Off topic, but what's the UX decision behind showing the line/column in the dialog. Removing it would be much better as we're already at the correct position, then only showing it in the footer area. That would also allow the tooltip size to expand properly and also allow the footer solution to use a float which would let the text content to wrap around the link.

steelbrain:
We have `remove link on the first line of bubble` in our todo. showing it in the bottom panel is a good thing IMO

blakeembrey:
@steelbrain Yes, keep it in the bottom - I was just suggesting a `float: right` to allow the lint text to flow around the link (using flexbox doesn't solve this). Some messages get kind of long right now, so I'll look into this quickly.

JohnMurga:
With the latest version the error is back for me ...
untitled

I worked around it just now with the following lines in my styles.less :+1: 
```
.linter-message-item {
  height:1.8em;
  white-space:nowrap;
}
```

blakeembrey:
@JohnMurga I'll look at replicating it, but both those rules are _very bad_ as it will break any unusual layout. However, you could say `nowrap` on the link though, that's a simple solution (but I'll try to replicate this first).

blakeembrey:
@JohnMurga Do you have a case I can replicate this?

JohnMurga:
@blakeembrey ...
I see this on a big screen.

Try putting :
```
require('web-frame').setZoomFactor(1.08);
```
In your init.coffee

As an aside, last night's fix was just taken from a previous version.

I only need :
```
.linter-message-item {
  white-space:nowrap;
}
```

To fix the issue for me.

tomitrescak:
Looks weird in the latest version ;( The original PR looked quite nice.

screen shot 2015-10-01 at 14 38 09

steelbrain:
@tomitrescak the `TS` icon in the bubble looked better in my theme, no idea why it looks like that in yours.

JohnMurga:
@tomitrescak I see you also have the wrapping problem ...

Did you try "File / Open your stylesheet" and adding the LESS I posted earlier ?

steelbrain:
@blakeembrey just a side note, your solution works beautifully on single line messages, but it beaks multi-line ones.

tomitrescak:
Yes I tried, but as mentions @steelbrain I need my multiline messages ;(( 

The multiline messages as solved in this PR still do not appear correctly.

blakeembrey:
@steelbrain Yes, you're right, it was only for his specific use-case. I'll look at how this looks with other themes, there is likely some defaults I didn't think about at work here.

blakeembrey:
@tomitrescak The badge looks like that in my theme too, maybe good to open that separately.

blakeembrey:
@steelbrain I might need more information on the multiline display, I can't seem to get it to function (E.g. it is splitting, but I can't display more lines). Why don't we just treat it like everything else and the above solution in CSS? We could make the overflow hidden by default using CSS.

steelbrain:
@blakeembrey The point of the multi line message is that only first line is showed in the beginning, then we add a class that shows all lines of it, currently you have to map that command to a key and when that command is pressed, all multi-line messages are expanded, I don't like it either, that's why I am making the expand icon work for individual messages as soon as I have enough time.

steelbrain:
Map `linter:expand-multiline-messages` to a key, doesn't work from command pallete

blakeembrey:
That's pretty confusing, I kind of expected to be able to click on the message and have it work - maybe look at making that change?

blakeembrey:
Also, personally, I wouldn't be doing the current hack and trying to manipulate HTML - I'd stick with text for multiline support.

steelbrain:
I didn't add support for these, it was contributed by the community :stuck_out_tongue:

steelbrain:
Also about that click to expand, here's the bug https://github.com/atom-community/linter/issues/914

blakeembrey:
@steelbrain I fixed some of the rendering issues I noticed. As for multiline, I did make some changes for clicking but removed it for the PR as I'm not sure the Atom etiquette for removing DOM listeners automatically - does it work with disposables or should I create event listeners on the specific elements?

steelbrain:
Atom does provide some libraries to convert DOM events into disposables, but simple `{add, remove}EventListener` should work just fine. Yeah, we are gonna make them element specific.

PR #903, opened 2015-09-25
Disable path in bubble and improve link/text flow

blakeembrey:
Here's an alternative keeping the text inline with the flow (meh):

image

Here's the current PR (same behaviour as before, but text will continue flowing below):

image

And before (with bubble link):

image


steelbrain:
Beautiful. anoter changelog entry would be appreciated

blakeembrey:
@steelbrain There is one in there, you might have been looking at my initial commit?

steelbrain:
I must be over caffeinated :sweat_smile: LGTM Thanks!

steelbrain:
Thanks for your awesome contributions to this project :clap:

PR #917, opened 2015-10-01
Enable clicking multiline messages + other tweaks

blakeembrey:
**Now:**

image

image

image

image

**Before:**

image

image

### DIFF
--- a/lib/ui/message-element.js
+++ b/lib/ui/message-element.js
@@ -65,24 +65,20 @@ export class Message extends HTMLElement {
     return el
   }
   static getMessage(message, includeLink) {
-    if (
-      message.multiline ||
-      NewLine.test(message.html) ||
-      NewLine.test(message.text)
-    ) {
+    if (message.multiline || NewLine.test(message.text)) {
       return Message.getMultiLineMessage(message, includeLink)
     }
 
     const el = document.createElement('span')
-    const messageEl = document.createElement('span')
+    const messageEl = document.createElement('linter-message-line')
 
     el.className = 'linter-message-item'
+
+    el.appendChild(messageEl)
 
     if (includeLink && message.filePath) {
       el.appendChild(Message.getLink(message))
     }
-
-    el.appendChild(messageEl)
 
     if (message.html && typeof message.html !== 'string') {
       messageEl.appendChild(message.html.cloneNode(true))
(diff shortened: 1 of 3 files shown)


blakeembrey:
@steelbrain Two remaining issues:

1. The first line of a multiline is now inconsistent with the normal line text when wrapping occurs. Personally, I would like to make the tweak to make the link for single line text to also have links flow inline with the text properly (I was previously trying to avoid that because of the previous issues opened, but I think it's preferable).
2. Why would someone use the `multiline` flag? Is it used? As it is, it's a little broken as we assume it is certainly multiline if they say it is, but when it's not the experience will break. In contrast, we are already checking if it's multiline. So the fix here for that issue would be to instead of the check, which results in `multiline` being a redundant flag. Does that make sense?

lib/ui/message-element.js
```diff
@@ -96,25 +92,35 @@ export class Message extends HTMLElement {
...
+
+      // Render the link in the "title" line.
+      if (index === 0 && includeLink && message.filePath) {
+        messageEl.appendChild(Message.getLink(message))
+      }
+    })
 
     container.appendChild(messageEl)
 
+    messageEl.addEventListener('click', function(e) {
```
steelbrain: How about you add this listener on line 109? You have the element there, you can assign it to a var, add the listener, then append child. That way you won't have to do the tagName check and it'll work even when we make changes to the element in the future.

steelbrain:
About Point `1`, do whatever pleases you. About point `2` `ide-hasksell` does I think, the newline check is just a test in case they missed it. but I get your point, maybe we should depreciate that flag?

lib/ui/message-element.js
blakeembrey: That could work, but I wanted to make all lines open and close, not just the first.

steelbrain: Wait, did I write that text above? Disregard that, I feel like I am out of my mind these days. It's perfectly fine to make all the lines clickable. It **could** conflict with it when we'll make the text selectable in the future, but we'll solve it then instead of now.

blakeembrey:
@steelbrain Made the changes for 1. For 2, yeah, I think it should be deprecated until it's properly thought through. If it's going to be deprecated, I'll make another PR for that and we remove some of the duplicate code.

steelbrain:
Tested it locally, looks good to me :clap:

PR #919, opened 2015-10-02
use info as a style key, as well as trace

devoncarew:
Our plugin generates errors, warnings, and info level analysis issues. We'd like to be able to call them that in the UI; currently we use `Error`, `Warning`, and `Trace` in order to get css styling in the editor for info level events. This PR adds `info` as a synonym for `trace`.

screen shot 2015-10-01 at 11 32 53 pm


steelbrain:
Thanks

PR #920, opened 2015-10-02
remove css top offset

devoncarew:
Remove a css `top:` offset. This seemed to be causing the linter status line contribution to be slightly off-center. I'm not sure if this is intentional or an artifact of some refactoring -

before:
screen shot 2015-10-01 at 11 54 14 pm

screen shot 2015-10-02 at 12 01 19 am

after:
screen shot 2015-10-01 at 11 54 24 pm


steelbrain:
:tada:

PR #921, opened 2015-10-02
made the text in the bottom panel selectable

devoncarew:
Fix https://github.com/atom-community/linter/issues/909 - made the text in the bottom panel selectable.


steelbrain:
Thank you so much :+1:

PR #927, opened 2015-10-03
Fix linter gutter alignment cross-theme/zoom

blakeembrey:
**Before:**

image

image

**After:**

image

image

Summary: Uses flexbox and ems to align properly at any scale.

Edit: Updated before images.


steelbrain:
cc @peduxe :wink:

steelbrain:
LGTM Merge at will.

PR #929, opened 2015-10-04
Improve error panel sizing behaviour

### LINKED ISSUE #928: Resizing Error Panel overides current auto-adjust behaviour
When using the new resize ability on the bottom error panel, if a size other than the default is chosen, then the panel remains persistent and stays on even when no messages are displayed. So when you move out of its current scope (onto a new line if set to only show messages from the current line, onto a new file if set to display from current file only...) it stays fixed. Unlike the case when left to the default which allows for auto-adjusting its size.

linter_panel_resize

I think the optimum behaviour would be to still auto-size in all cases, as is currently done for default, but that the resize should act as the new upper limit (whereas the default is effectively 'unlimited'). So the panel would disappear whenever there are no error messages and still auto adjust if the user, for example, sets the size to ~4 lines and only 1 error is present.

blakeembrey:
Adds two options for controlling panel size - `errorPanelHeight` and `alwaysTakeMinimumSpace`. The error panel height can be manually configured, or automatically updated by using the drag feature  (results in persistence of height). The minimum space option is a toggle which causes the height of the panel to equal `min(height option, messages height)` (vs `height option`). This patch also closes an issue where the linter panel would stay open with no messages visible.

Closes #928

**Empty Linter Panel Issue:**

image

Edit: Had to modify tests to set-up visibility before checking (not visible by default when there's no error messages anymore).

### DIFF
--- a/lib/ui/bottom-panel.js
+++ b/lib/ui/bottom-panel.js
@@ -8,54 +8,88 @@ export class BottomPanel {
   constructor(scope) {
     this.subscriptions = new CompositeDisposable
     this.element = document.createElement('linter-panel') // TODO(steelbrain): Make this a `div`
-    this.element.tabIndex = "-1"
+    this.element.tabIndex = '-1'
+    this.messagesElement = document.createElement('div')
     this.panel = atom.workspace.addBottomPanel({item: this.element, visible: false, priority: 500})
     this.visibility = false
+    this.visibleMessages = 0
+    this.alwaysTakeMinimumSpace = atom.config.get('linter.alwaysTakeMinimumSpace')
+    this.errorPanelHeight = atom.config.get('linter.errorPanelHeight')
+    this.configVisibility = atom.config.get('linter.showErrorPanel')
     this.scope = scope
     this.messages = new Map()
 
-    this.subscriptions.add(atom.config.observe('linter.showErrorPanel', value => {
-      this.configVisibility = value
-      this.setVisibility(true)
-    }))
-    this.subscriptions.add(atom.workspace.observeActivePaneItem(paneItem => {
-      this.paneVisibility = paneItem === atom.workspace.getActiveTextEditor()
-      this.setVisibility(true)
+    // Keep messages contained to measure height.
+    this.element.appendChild(this.messagesElement)
+
+    this.subscriptions.add(atom.config.onDidChange('linter.alwaysTakeMinimumSpace', ({newValue, oldValue}) => {
+      this.alwaysTakeMinimumSpace = newValue
+      this.updateHeight()
     }))
 
-    Interact(this.element).resizable({edges: { top: true }})
-      .on('resizemove', function (event) {
-        var target = event.target;
-        if (event.rect.height < 25) {
-          if (event.rect.height < 1) {
-            target.style.width = target.style.height = null
-          } else return // No-Op
-        } else {
-          target.style.width  = event.rect.width + 'px'
-          target.style.height = event.rect.height + 'px'
-        }
+    this.subscriptions.add(atom.config.onDidChange('linter.errorPanelHeight', ({newValue, oldValue}) => {
+      this.errorPanelHeight = newValue
+      this.updateHeight()
+    }))
+
+    this.subscriptions.add(atom.config.onDidChange('linter.showErrorPanel', ({newValue, oldValue}) => {
+      this.configVisibility = newValue
+      this.updateVisibility()
+    }))
+
+    this.subscriptions.add(atom.workspace.observeActivePaneItem(paneItem => {
+      this.paneVisibility = paneItem === atom.workspace.getActiveTextEditor()
+      this.updateVisibility()
+    }))
+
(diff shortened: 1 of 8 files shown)


lib/ui/bottom-panel.js
```diff
@@ -8,65 +8,98 @@ export class BottomPanel {
...
 
+    // Keep messages contained to measure height.
+    this.element.appendChild(this.messagesElement)
+
+    this.subscriptions.add(atom.config.observe('linter.alwaysTakeMinimumSpace', config => {
+      this.takeMinimumHeight = config
+      this.updateHeight()
+    }))
+
+    this.subscriptions.add(atom.config.observe('linter.errorPanelHeight', height => {
```
steelbrain: use `onDidChange` instead of observe here 'cause you're causing `updateHeight` just a few lines up.

lib/ui/bottom-panel.js
```diff
@@ -8,65 +8,98 @@ export class BottomPanel {
   constructor(scope) {
     this.subscriptions = new CompositeDisposable
     this.element = document.createElement('linter-panel') // TODO(steelbrain): Make this a `div`
-    this.element.tabIndex = "-1"
+    this.element.tabIndex = '-1'
+    this.messagesElement = document.createElement('div')
```
steelbrain: Why do we need this element? Is it for styling purposes?

blakeembrey: Purely for calculating the height, actually.

lib/ui/bottom-panel.js
blakeembrey: Why would that help here? They are caused by different changes.

lib/ui/bottom-panel.js
steelbrain: Fine by me.

lib/ui/bottom-panel.js
steelbrain: observe is called instantly, `onDidChange` is called when something changes, so this way, the same function is getting called twice, You will have to make either of them an `onDidChange`.

blakeembrey: Gotcha, now I understand. I don't think either need to be observed technically then, I'll play around with it.

blakeembrey: Oh, they were both doing it to get the values. I can move it about and make it init from config and both use `onDidChange` instead.

steelbrain: How about you use `onDidChange` on the one above and `observe` in the one below, that way both will be triggered on change, and when the last one finally calls it, it'll have correct values.

blakeembrey: I'd prefer none making unnecessary calls, so I've made them all `onDidChange`.

steelbrain: If you would follow what I said above, there would be no unnecessary calls :sweat_smile: that's why I was suggesting that

blakeembrey: No, they were all already unnecessary actually. You only need the one `updateVisibility` and I've fixed that now. And actually, I don't think you need that since `setMessages` calls `updateVisibility` which is when you need it.

blakeembrey: Actually, ignore that - it's the only required one.

lib/ui/bottom-panel.js
```diff
@@ -8,65 +8,101 @@ export class BottomPanel {
...
+        atom.config.set('linter.errorPanelHeight', event.target.clientHeight)
       })
   }
   refresh(scope) {
     this.scope = scope
+    this.visibleMessages = 0
+
     for (let message of this.messages) {
-      message[1].updateVisibility(scope)
+      const visible = message[1].updateVisibility(scope)
```
steelbrain: Is assigning a var here really necessary? Why not just use the value like
```js
if (message[1].updateVisibility(scope)){
  visibleMessages++
}
```

lib/ui/bottom-panel.js
```diff
@@ -8,65 +8,101 @@ export class BottomPanel {
...
       this.removeMessages(removed)
+
     for (let message of added) {
       const messageElement = Message.fromMessage(message)
-      this.element.appendChild(messageElement)
-      messageElement.updateVisibility(this.scope)
+      this.messagesElement.appendChild(messageElement)
+      const visible = messageElement.updateVisibility(this.scope)
       this.messages.set(message, messageElement)
+      if (visible) this.visibleMessages++
```
steelbrain: The same https://github.com/atom-community/linter/pull/929/files#r41109890 applies here

steelbrain:
Overall, Good work! :clap:

lib/ui/bottom-panel.js
blakeembrey: Readability. But I can change it.

blakeembrey:
@steelbrain Fixed the remaining couple of comments.

steelbrain:
LGTM. Thanks again :rose:

PR #945, opened 2015-10-09
Make copy to clipboard work in panel

### LINKED ISSUE #937: Copy selected message text to clipboard
Although https://github.com/atom-community/linter/pull/921 allows the selection of message text, apparently electron doesn't automatically handle copying it to the clipboard when a copy event is fired (Ctrl + C).

Please update linter to handle this case as selectable message text is of little use without being able to copy the error :wink:.

steelbrain:
Fixes #937

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/945)

### DIFF
--- a/lib/ui/bottom-panel.js
+++ b/lib/ui/bottom-panel.js
@@ -1,6 +1,7 @@
 'use babel'
 
 const Interact = require('interact.js')
+const Clipboard = require('clipboard')
 import {CompositeDisposable} from 'atom'
 import {Message} from './message-element'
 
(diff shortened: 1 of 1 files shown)


lib/ui/bottom-panel.js
```diff
@@ -1,6 +1,7 @@
 'use babel'
 
 const Interact = require('interact.js')
+const Clipboard = require('clipboard')
```
steelbrain: This is an electron module https://github.com/atom/electron/blob/master/docs/api/clipboard.md

PR #950, opened 2015-10-11
Sort messages in bottom panel

### LINKED ISSUE #833: `Linter: Next Error` jumps to seemingly random error

### LINKED ISSUE #865: Sort messages
Edit by @steelbrain
Things to do:
  - [x] Sort by file
  - [ ] Sort by line

The messages in the panel should be sorted by line number (then column, then ?). This will make it so that when multiple linter providers send a message to `linter` things are not completely out of order.

It should also fix an issue where messages from the same provider are out of order, which I think is caused by new messages being added at the bottom while old ones are not touched.

steelbrain:
We can then use the same sorted messages in commands to jump to next message in asc order.

ToDo:
  - [x] Make it keep messages of different files separate

Fixes #865 
Fixes #833

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/950)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Major speed improvements
 * Make ctrl-c work on bottom panel
 * Fix certain scenarios where inline bubbles would be placed incorrectly (Bubbles no longer follow the cursor, they re-use markers from underlines)
+* Implement sorted messages (comes at a performance penality, can be turned off in configuration)
 
 ## 1.8.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Major speed improvements
 * Make ctrl-c work on bottom panel
 * Fix certain scenarios where inline bubbles would be placed incorrectly (Bubbles no longer follow the cursor, they re-use markers from underlines)
-* Implement sorted messages (comes at a performance penality, can be turned off in configuration)
+* Implement sorted messages (comes at a performance penalty, can be turned off in configuration)
 
 ## 1.8.1
 
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -17,6 +17,11 @@ module.exports =
       items:
         type: 'string'
       order: 2
+    sortMessages:
+      description: 'This comes at a performance penalty'
+      type: 'boolean'
+      default: true
+      order: 2
 
     showErrorInline:
       title: 'Show Inline Tooltips'
--- a/lib/ui/bottom-panel.js
+++ b/lib/ui/bottom-panel.js
@@ -18,6 +18,7 @@ export default class BottomPanel {
     this.errorPanelHeight = atom.config.get('linter.errorPanelHeight')
     this.configVisibility = atom.config.get('linter.showErrorPanel')
     this.scope = scope
+    this.sortedMessages = []
     this.messages = new Map()
 
     // Keep messages contained to measure height.
@@ -36,6 +37,10 @@ export default class BottomPanel {
     this.subscriptions.add(atom.config.onDidChange('linter.showErrorPanel', ({newValue}) => {
       this.configVisibility = newValue
       this.updateVisibility()
+    }))
+
+    this.subscriptions.add(atom.config.observe('linter.sortMessages', sortMessages => {
+      this.sortMessages = sortMessages
     }))
 
     this.subscriptions.add(atom.workspace.observeActivePaneItem(paneItem => {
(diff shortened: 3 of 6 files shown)


steelbrain:
We have two options here.

 - Use the current approach which is like a performance overkill IMO
 - Allow sorting only via file path and store messages from each file path in separate html elements, show/hide them on pane change and scope change. This way all messages from one path would be showed in one place, but they won't be sorted via ranges and won't hurt performance at all.

I am not sure which one we should choose. Suggestions?

/cc @k2b6s9j @blakeembrey

steelbrain:
After some discussion with @Arcanemagus we found a new way to solve this. What if we store the messages based on their filePaths in different containers. Because the cost of iteration over that container is ridiculously high, we could just detach the elements in that and that container only, sort them after merging the new messages and sort and append them. That would not only minimize the iteration overhead that at the moment is `number of new messages x number of previous messages` to `(number of previous messages + number of new messages) x 2` but will also reduce DOM detach and reattach operations.

SpainTrain:
Are there numbers on how long it takes to perform these ops given a realistic `n` (and what do we consider an extremely high `n`? 50?)?  If the speed difference is imperceptible, then :+1: to whatever is most simple and easiest to maintain.  If option 3 is noticeably faster, then :+1: to option 3.  Also, for implementing sorted insertion, it may make sense to just use an existing lib, e.g. https://www.npmjs.com/package/sorted-array

Related question:  Does atom collect telemetry, and/or allow plugins to gather telemetry?  It would be incredibly useful and interesting for dev purposes to know how linters are being used in the wild.  For example, we could answer the question of what is average `n` and extreme edge case `n` (where `n` is # lint violations).


---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[lib/main.coffee, line 21 \[r1\]](https://reviewable.io:443/reviews/atom-community/linter/950#-K0Rv_xbRR5qBelVFjZC)** ([raw file](https://github.com/atom-community/linter/blob/d8541be58b02e6a15f1cf1c51b08590c03653f7b/lib/main.coffee#L21)):</sup>
nit: This should probably also describe the feature ;-)  e.g., 'Sort lint messages in the bottom panel (caveat: incurs performance penalty)'


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/atom-community/linter/950)

steelbrain:
@SpainTrain Well I know that the nuclide collects analytics, atom has a metrics package too so I think it's okay for us to collect anonymous statistics.

SpainTrain:
:+1: A PR someday

blakeembrey:
I think extremely high should be considered 1000+. I easily hit it when I open a different project with the standard linter enabled (E.g. every semicolon, braces position, etc.). Also, I'm not 100% sure about having a toggle for this - I think it should be all in or all out, as I'd definitely enable it but it doesn't mean I want it to blow up on the occasional file/project. It could always be a "beta" toggle that we remove later.

As for the sort algorithm, I haven't had the time to review absolutely but off the top of my head we should be able to decent performance by keeping each file in it's own object/array and sorting those before we pass them around. We should be able to implement an O(n) sort there. It might even be possible to abuse the way objects work in V8 - it keeps numeric keys in order already. Then each message could be stored in a linked list or array at that key index.

I could try an implementation like this later in the week, so feel free to implement it however now too.

Arcanemagus:
@blakeembrey I suggested @steelbrain look into implementing a Red-Black binary tree, which should get us into O(log _n_) territory.

We'll see how hard it is to actually implement though.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Major speed improvements
 * Make ctrl-c work on bottom panel
 * Fix certain scenarios where inline bubbles would be placed incorrectly (Bubbles no longer follow the cursor, they re-use markers from underlines)
+* Implement sorted messages (comes at a performance penality, can be turned off in configuration)
```
SpainTrain: nit: "penalty"

steelbrain: :stuck_out_tongue:

SpainTrain:
> keeping each file in it's own object/array and sorting those

That's where I got the 50 number from (though should still probably be higher).  Agree that whole project would be a much larger potential total violations.

What is relevance/meaning of sorting at the project-level?  Wouldn't the most useful ordering be "group by file, sort by line"?  Also, is it reasonable to apply some sane limit to the number of items actually rendered into the bottom panel, with a "plus 128,321 more" or similar at the bottom?

SpainTrain:
lol, nevermind, I have been set straight with the knowledge that _single files_ sometimes have 1k+ linting errors :mask:

blakeembrey:
> What is relevance/meaning of sorting at the project-level? Wouldn't the most useful ordering be "group by file, sort by line"? Also, is it reasonable to apply some sane limit to the number of items actually rendered into the bottom panel, with a "plus 128,321 more" or similar at the bottom?

Yes, I would expect group by file and sort by line too which is why the filePaths containers made sense to me. As for the limit, I think you're right there too. It definitely needs to be capped for rendering, maybe even just the first 50 and I like the idea of an extra text item that says "x more" and the tab in the bottom can also render the correct number.

steelbrain:
I think I am gonna close this one and implement file-specific containers for now in a separate PR. Range based sorting will come later on.

PR #957, opened 2015-10-16
Sort messages via file containers

steelbrain:
This should make the internals a bit cleaner, and it has no performance regressions at all

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/957)


steelbrain:
Alright, the implementation is done. Also fixed some bugs that were introduced by recent PRs

PR #979, opened 2015-10-30
Fix a styling issue in bottom tabs

### LINKED ISSUE #978: Styling issue when status bar tabs are hidden
The following styles in `linter-plus.less` do not work when hiding tabs in the status bar. Namely the "line" and "project" tabs.

```less
&:first-of-type {
  border-top-left-radius: @component-border-radius;
  border-bottom-left-radius: @component-border-radius;
}
&:last-of-type {
  margin-right: .6em;
  border-top-right-radius: @component-border-radius;
  border-bottom-right-radius: @component-border-radius;
}
```

This happens because you hide them by adding an attribute `hidden` to the DOM elements rather than removing them from the DOM. This results in the following (when both "line" and "project" tabs are hidden):

screenshot from 2015-10-30 01-58-36

As you can see, the button is missing both the round corners and the right margin. The lack of a right margin looks rather bad, in my opinion.

This seems tricky to fix by only messing with the styles. I believe a good way to fix this would be to place the tabs inside a container and style the container with the outer border (with radius) and right margin.

Thank you for your amazing work so far! :smile:

steelbrain:
Fixes #978

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/979)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Upcoming
+
+* Fix a minor style issue in bottom tabs
+
 ## 1.9.1
 
 * Add `ignoreMatchedFiles` config
--- a/styles/linter-plus.less
+++ b/styles/linter-plus.less
@@ -111,15 +111,6 @@ linter-bottom-tab {
   cursor: pointer;
   vertical-align: middle;
   position: relative;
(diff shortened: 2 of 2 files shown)


steelbrain:
Before
Before
After
After

Issue #964, opened 2015-10-22
A better self-API


steelbrain:
Here's what I have in mind
```coffee
consumeLinter: (linterRegistryHelper) ->
  linterAgent = linterRegistryHelper.registerLinter({
    name: 'some-linter'
  })
  linterAgent.setMessages([...messages...])
  linterAgent.deleteMessages()

  @subscriptions.add(linterAgent) # or
  linterAgent.dispose() # automatically deletes everything of this agent
```
^ It should make a lot of things simpler. We have been told that our self-API isn't as beautiful so now that we're done with high priority things, I think this is the best time to work on it

cc @radicaled @devoncarew @david-driscoll @russpowers

^ authors of packages using self-API. I did a quick github code search if I've missed somebody, please ping them on here

CHANGELOG
 - Removed `scope` from the API. I haven't seen a PUSH style linter that is per file yet, If you have, let me know.

kepler0:
And I just started using this for atom-build...

This does look better though.

kepler0:
What about methods to find all messages for packages like `minimap-linter`?

steelbrain:
@kepler0 those methods are gonna stay as-is. Those methods are pretty fine, it's just that I was told recently that linter doesn't support PUSH mechanism and only supports PULL mechanism to which I replied with self-API and they said it wasn't documented anywhere and they were right.

Packages using self-API for non-provider purposes have too broad coverage, we can not wrap all of that, but we can polish our provider self-api because that's what this package is for.

kepler0:
Awesome. Just making sure that these won't be the only methods exposed.

devoncarew:
So, my drive-by comment -

We've been able to build what we need to on the current self API. This iteration looks good; happy to consume that as well :) We would appreciate a period of overlap - where we could move to the new API, but the current would would continue to be available.

steelbrain:
@devoncarew Absolutely, even when we roll that API out, the current packages will continue working 100%

steelbrain:
I'll start working on this now.

kepler0:
HYPE!

steelbrain:
Added in #986

PR #986, opened 2015-11-03
WIP: A better self API

steelbrain:
See #964 for API

ToDo
 - [x] Implement API
 - [x] Write specifications
 - [x] Update wiki

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/986)


steelbrain:
The API implementation is finished. All we need now is some specs and wiki change.

steelbrain:
Specs added, wiki updated.

PR #990, opened 2015-11-05
Better name on the `linter:togglePanel` command

KnutHelland:
`linter:togglePanel` is not a good command name, it should be `linter:toggle-panel`.

It should be consistent with command name conventions.

### DIFF
--- a/lib/commands.coffee
+++ b/lib/commands.coffee
@@ -7,7 +7,7 @@ class Commands
       'linter:next-error': => @nextError()
       'linter:previous-error': => @previousError()
       'linter:toggle': => @toggleLinter()
-      'linter:togglePanel': => @togglePanel()
+      'linter:toggle-panel': => @togglePanel()
       'linter:set-bubble-transparent': => @setBubbleTransparent()
       'linter:expand-multiline-messages': => @expandMultilineMessages()
       'linter:lint': => @lint()


review-ninja:
<a href="https://app.review.ninja/atom-community/linter/pull/990" target="_blank">ReviewNinja

gdi2290:
@review-ninja review my stuff

steelbrain:
@KnutHelland This would be a key mapping breaking change, I don't think it's worth it.

steelbrain:
I think I would keep this PR open for when we're bumping a major version

KnutHelland:
Yes, I was about to suggest the same thing. You could also alternatively have them both, but let the old event trigger an deprecation warning until it is removed in the next major version.

I don't think it is wise to ignore this thing. I think it is quite bad to break this naming convention. When I was going to map a key to this command earlier today, I had to look up the source code for finding out what the real name for this event really was. Because it looks like it should be `linter:toggle-panel` in the command palette, but there is no space between `toggle` and `panel`, so I should have understood it. But this is the first command I've ever seen that breaks this convention so far, after having used Atom for half a year.

screen shot 2015-11-05 at 15 56 34

steelbrain:
Ref: https://github.com/atom/atom/issues/9456

jeancroy:
One thing to consider, if this is using fuzzy matching camelCase and kebab-case are treated the same (aka the `p` of pannel will be promoted). Also adding a character is a free operation, you can continue to not type it and it'll find it's way.

steelbrain:
@KnutHelland I've added this command in the new linter ui package, I've made sure to use the name you suggested, see https://github.com/AtomLinter/linter-ui-default/pull/23

PR #991, opened 2015-11-06
Auto updates

### LINKED ISSUE #960: Auto-Updates
I am thinking of an option, that's enabled by default, that checks for updates for linter package and installs them.

What do you think guys?
cc @k2b6s9j @SpainTrain @Arcanemagus @atom-community/linter-contributors

steelbrain:
Fixes #960

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/991)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Upcoming
 
 * A better self-API for PUSH style providers. Check the wiki for more information.
+* Linter will now automatically try to update if one is available, you can disable it in settings.
 
 ## 1.10.0
 
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -132,10 +132,10 @@ module.exports =
       order: 5
 
   activate: (@state) ->
-    if atom.config.get('linter.autoUpdate')
-      require('atom-package-upgrader').upgrade('linter')
     LinterPlus = require('./linter.coffee')
     @instance = new LinterPlus state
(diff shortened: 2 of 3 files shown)


kepler0:
The implementation looks great. I still don't know that I support the idea though...

blakeembrey:
@steelbrain I just wanted to chime in too, but I would personally prefer not to do this in general. If people aren't updating, there's probably some reason. Also, if I look over the code correctly, this is only handling Atom startup? What about when the user has a long running editor open? I normally have mine open and running for weeks, but the updates come to me as usual still.

steelbrain:
@blakeembrey People usually don't upgrade because they don't see the upgrade notification or don't pay attention to it. There are times when we introduce a bug accidentally and then immediately push a release that fixes that, this will make sure the users have the latest version of linter installed. and because we don't break the provider API, the latest version never hurts.

steelbrain:
@blakeembrey I've updated the package to now query for updates each 12 hours

kepler0:
There should be an informal study if this behavior would influence users who use Atom heavily.

Perhaps start with a survey on the Atom Slack.

kepler0:
I'd also like to know the opinion of this from larger project which rely on Linter (OmniShard, Ionide, Nuclide, etc.)

steelbrain:
After some discussion with @kepler0 I've decided not to do this in our package, we'll wait for core to do this instead

PR #995, opened 2015-11-07
WIP: Import UI as a service

### LINKED ISSUE #865: Sort messages

### LINKED ISSUE #885: Trigger tooltips on hover
That's how IDEs do it. and Atom is more than just an editor right?
I would love to accept a PR for this as I am mostly busy in personal work these days.

### LINKED ISSUE #985: Import UI as a service
We have been considering this from day 1. Now that the other things are done, we can focus on doing this. The idea is to separate the bottom panel UI components into a separate package. Export a service, consume that service in that package and show those UI components. It'll not only make the core more `modular` but will also allow the other package authors to get most out of this package.
Everybody can then tweak linter to match their taste.
Here's the API I have in mind
```js
class Provider {
  constructor()
  activate(): void
  didCalculateMessages({added, removed, messages}): void
  didBeginLinting(linter, filePath): void
  didFinishLinting(linter, filePath): void
  dispose(): void
}
```

#### Changelog
  * Remove `set messages` and `set count`, replace with `update(messages: Array<Message>)`
  * Convert UI provider into a class
  * Add events to the API and rename update method
  * Rename `initialize` to `activate` (Suggestion by Arcanemagus )
  * Remove parameter to activate method

### LINKED ISSUE #998: Linter Toggle fails to remove bubble gutter on left
When first opening a file from the tree view, the linter is invoked (say, rubocop on a foo.rb file) and a gutter is added to the left hand side of the content window. If you then use linter toggle to try to dismiss it, the panel is hidden but the gutter with the line 'bubbles' remains. You can toggle linter on and then off again and the gutter will get removed, it just never works the first time, in my experience.

linter-toggle-bug-01
linter-toggle-bug-02
linter-toggle-bug-03

### LINKED ISSUE #1004: Linter not catching up until saving
I've found in files with 1k+ lines the linter can get to the point where it's too slow to lint inline and as I type I get errors that just "stick".
For example this error didn't go away at all until I saved even though it picked up other errors that I fixed in a few lines below it.

screen shot 2015-11-15 at 2 42 33 pm

Edit: This file has over 1900 lines and I don't get the same issue with smaller files.

### LINKED ISSUE #1007: Very unpredictable and buggy behaviour when file is open in multiple panes
Hi,

Sorry for an unspecific title and not much information. I want to give a brief description now, and _if this isn't an already known issue, I'll investigate, update title and post more information on this issue_.

I've used the linter for months now, and there is a bug annoying me all the time. If I have the same file open in multiple panes (which I do a lot. Split screen into the same file), it behaves very buggy and unpredictable. Often new errors are displayed only in one of the panes, but if I fix the error, I have to go back to the other pane and save the file from there before it gets linted again so the error disappears.

This happens with both `linter-eslint` and `linter-less`, so it is probably a problem with `linter` itself.

steelbrain:
Fixes #985
Fixes #1007
Fixes #998
Fixes #885
Fixes #865 
Fixes #1004

If somebody needs a reason for this, they can ask me on slack

ToDo
 - [x] Add UI Registry
 - [x] Import UI Service
 - [x] Finalize a spec along the way
 - [x] Add specs confirming it works
 - [x] Split ui into `linter-ui-default` package
 - [x] Install `linter-ui-default` by default
 - [x] Remove ui folder completely (All done except panel, will delete when resize is implemented in panel)


steelbrain:
This seems to have stalled, I'll work on it now

steelbrain:
The API is finalized and passes every scenario I can think of right now. The only thing left is specs and moving styles into the ui package

steelbrain:
I've tweaked and removed some specs. and they seem to be passing now!

steelbrain:
YES YES YES! HELL YES!
I've brought down the message registry initialization and diffing time of 5k messages down from 25-30ms to 8-11ms!

steelbrain:
so I've rewritten the message registry entirely, the results are awesome! It now takes 4-10ms to calculate diff between 5k messages. I've also enabled validation of messages/linters in dev mode only.

:bomb: :bomb: :bomb: :bomb: :bomb: :bomb: :bomb: :bomb:

steelbrain:
Gonna merge this since this is done, but will hold off merge until `linter-ui-default` package is finished and published

florianb:
Good work! :cake: & :coffee:

steelbrain:
Added the wiki page here https://github.com/atom-community/linter/wiki/Linter-UI-API

PR #1017, opened 2015-12-13
Use buffer as key to store messages

steelbrain:
It was added in a PR, reverted accidently, now adding again

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/1017)

### DIFF
--- a/lib/message-registry.js
+++ b/lib/message-registry.js
@@ -34,19 +34,23 @@ class MessageRegistry {
     }
     try {
       Validate.messages(messages, linter)
-    } catch (e) { return Helpers.error(e) }
-    messages = messages.filter(i => this.ignoredMessageTypes.indexOf(i.type) === -1)
+    } catch (e) {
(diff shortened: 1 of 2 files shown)


steelbrain:
Travis seems to be acting weird, specs pass locally for me
```
~/.a/p/linter ❯❯❯ apm test                                                                                 steelbrain/message-registry-enhancements
Window load time: 630ms
................................................................................................
Finished in 4.429 seconds
96 tests, 240 assertions, 0 failures, 0 skipped

```

Arcanemagus:
The build has been failing since https://github.com/atom-community/linter/pull/1012 was merged in, all subsequent builds have failed.

Something is different between your local copy and Travis since it has one additional assertion compared to your output.

steelbrain:
I wonder how it is possible since my local copy and travis share the same specs folder. I'm using the latest apm by the way

PR #1018, opened 2015-12-13
Make single buffer work across multiple panes

### LINKED ISSUE #1007: Very unpredictable and buggy behaviour when file is open in multiple panes

steelbrain:
Fixes #1007

ToDo
 - [x] Add spcs

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/1018)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Upcoming
 
+* Fix linter message synchronization between buffers (See #1007)
 * Remove `Linter.getEditorLinterByPath` of self API (GitHub code search says nobody was using it)
 
 ## 1.11.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Upcoming
+
+* Remove `Linter.getEditorLinterByPath` of self API (GitHub code search says nobody was using it)
+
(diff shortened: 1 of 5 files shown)


steelbrain:
@Arcanemagus what's the status on your machine? It's still passing for me and failing on travis

steelbrain:
The specs failing are not related to this PR in any way so merging in

Arcanemagus:
Whoops, just now saw this. Status on my machine (after manual creation of `C:\tmp`):
image

steelbrain:
Thanks for showing me that! I'll be improving the specs once I'm done with this major improvement of message-registry I'm working on

PR #1019, opened 2015-12-13
Add a .gitattributes file

Arcanemagus:
This is set to force LF EOL matching the .editorconfig.


steelbrain:
LGTM, Merge at will

PR #1020, opened 2015-12-13
Add AppVeyor config

Arcanemagus:
Add a simple AppVeyor configuration in preparation for enabling it to test Windows builds here.


steelbrain:
This will help us make sure that we don't break our specs on Windows as we progress, Thanks :clap:

Arcanemagus:
Huh, `atom -v` isn't showing anything, but at least `apm -v` shows us output.

PR #1021, opened 2015-12-13
Enhance message registry

steelbrain:
It's a lot more efficient than it used to be, here are some of the key points

 - It's optimized when messages are received from a linter for the first time
 - It's optimized when no messages change
 - It doesn't do anything unless necessary

ToDo
  - [ ] Make Delete work
  - [ ] Fix specs

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/1021)

PR #1023, opened 2015-12-13
Re-land Message Registry improvements

steelbrain:
Relands #1021
I had accidentally committed it without delete support.
Gonna implement it here

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/1023)


steelbrain:
A benchmark shows that our current message registry is faster than the one introduced in this PR, therefore I am abandoning this PR

PR #1024, opened 2015-12-13
Add message registry benchmarks

steelbrain:
Just like the title says

The way I run them is that I run atom in spec runner mode in package root, then open it's dev console and paste this 
```
require('/home/steel/.atom/packages/linter/benchmarks/message-registry.js')
```

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/1024)

### DIFF
--- a/benchmarks/common.js
+++ b/benchmarks/common.js
@@ -1,0 +1,20 @@
+'use babel'
+
+import {getMessage} from '../spec/common'
+
+export function timeNow() {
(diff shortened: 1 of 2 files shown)


steelbrain:
These benchmarks show that our message registry is as optimized as it could ever be (accidentally I may add). Even changing that `for...of` into a foreach results in a performance regression

PR #1027, opened 2015-12-16
Fix Specifications

steelbrain:
They were working for me and failing for others before, trying to make it work for all of us now

[Review on Reviewable](https://reviewable.io/reviews/atom-community/linter/1027)

### DIFF
--- a/spec/config-spec.coffee
+++ b/spec/config-spec.coffee
@@ -1,9 +1,13 @@
 describe 'Linter Config', ->
   linter = null
   {getLinter, getMessage} = require('./common')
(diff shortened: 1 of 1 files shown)


steelbrain:
One fixed, not sure how to fix the others

Arcanemagus:
Use `FS.unlinkSync(filePath)` instead of `CP.execSync("rm -f #{filePath}"` to get rid of a few potential issues on Windows (the only reason that didn't show up in the specs is because AppVeyor runs in PowerShell by default which emulates those commands).

PR #1029, opened 2015-12-16
Fix file path ignoring

Arcanemagus:
Change the default and fix the specs for Windows.


Arcanemagus:
Closing in favor of #1030.

PR #1089, opened 2016-02-28
Enhancements

steelbrain:
- [x] Rewrite the remaining coffee script files into ES6+
- [x] Require at least Atom 1.4.0
- [x] Move config to package manifest


steelbrain:
Gonna implement Fix API in a separate PR to keep this focused

steelbrain:
@Arcanemagus Anything you wanna change before this is merged?

florianb:
Good work! Moving towards ES seems to be trendy -- i'll push the transformation forwards for linter-javac, too.

I think i'll quickly add the support for `configSchema`. :tada:

steelbrain:
@florianb Nice, Thanks :clap:

steelbrain:
@florianb Also note that moving to config schema will break your package for people using Atom older than 1.4.0 so you might also want to bump the minimum version

florianb:
@steelbrain - thank you very much! :clap:

 I already searched for a documentation about `configSchema` and would have missed the version-bump definitely! :candy: :candy: :candy:

Arcanemagus:
Shouldn't the intentions stuff be moved to the UI package? Linter shouldn't be tying itself to any particular implementation here, we can have `linter-ui-default` provide the service and rely on `intentions`.

steelbrain:
You are absolutely right @Arcanemagus, if we do that, we'll be allowing UI package authors to provide a fix button on messages directly maybe if they prefer

Arcanemagus:
Bumping the min version to v1.4.0 should probably stay though, I've seen a few issues in the `linter-*` providers where some ES6 functionality is missing in old versions of Atom.

lib/validate.js
```diff
@@ -0,0 +1,108 @@
...
+
+  return true
+}
+
+export function linter(linter: Linter$Linter, indie: boolean = false): boolean {
+  let message
+
+  if (!linter || typeof linter !== 'object') {
+    message = 'Linter Provider must be an object'
+  } else if (linter.name && typeof linter.name !== 'string') {
```
Arcanemagus: Before there was an `else` on this clause making `linter.name` be `null` if this evaluated to false, will this be `undefined` now?

steelbrain: Yes it'll be undefined now, we previously executed validate all the time, so we expected stuff to be either defined or null. Now we only execute it in dev mode, which means we have adapted our codebase to work with undefines so null is no longer required

lib/validate.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+  let foundError = false
+
+  for (const entry of entries) {
+    let message
+
+    if (typeof entry.type !== 'string') {
+      message = 'Linter message.type must be a string'
+    } else {
+      if (typeof entry.html !== 'undefined') {
```
Arcanemagus: It probably doesn't make a performance difference in a modern JS engine, but it might be slightly faster to check for `entry.text` first since most of the linter providers use this message type.

Arcanemagus:
LGTM :+1:

PR #1093, opened 2016-03-03
Fix a bug in message registry

steelbrain:
I believe this to be the reason for https://github.com/atom-community/linter/issues/1090
and the crashes @Arcanemagus is experiencing

After this, there is no known bug in message-registry :tada: :tada: :tada: :tada:


Arcanemagus:
LGTM

PR #1097, opened 2016-03-14
Enhance eslint config

steelbrain:
Just like the title says


steelbrain:
@Arcanemagus ESLint is passing in Circle and failing in Travis, ideas?

florianb:
@steelbrain It seems like you're failing your own eslint-rules. And circle-ci seems not to run `eslint lib` anyway.

steelbrain:
@florianb Well CircleCI is running the specs, look for `npm run lint` in https://circleci.com/gh/atom-community/linter/40

steelbrain:
Turns out we were linting just the `lib` directory in CircleCI and the entire package folder in other CIs. I've fixed the ESLint issues in all js files in the package so this should be good now

florianb:
This happened since Travis and AppVeyor completely rely on the atom/ci-build-scripts which detect and linter-packages like eslint and then automagically invoke the failing lintings for the whole project.

.eslintrc.json
```diff
@@ -0,0 +1,35 @@
+{
+  "parser": "babel-eslint",
+  "extends": "airbnb",
```
Arcanemagus: Not `airbnb/base`?

.eslintrc.json
```diff
@@ -0,0 +1,35 @@
...
+    "react/jsx-no-bind": 0,
+    "react/jsx-closing-bracket-location": 0,
+    "react/no-deprecated": 0,
+    "react/jsx-boolean-value": 0,
+    "react/sort-comp": 0,
+    "prefer-template": 0,
+    "no-nested-ternary": 0,
+    "no-multi-spaces": 0
+  },
+  "globals": {
```
Arcanemagus: All of these except `atom` should be in a file specifically for the specs, those globals aren't defined in general Atom.

Arcanemagus:
Are all of those rule overrides really necessary? I'd be a bit surprised...

.eslintrc.json
```diff
@@ -0,0 +1,35 @@
...
+    "react/prefer-stateless-function": 0,
+    "react/prop-types": 0,
+    "react/jsx-no-bind": 0,
+    "react/jsx-closing-bracket-location": 0,
+    "react/no-deprecated": 0,
+    "react/jsx-boolean-value": 0,
+    "react/sort-comp": 0,
+    "prefer-template": 0,
+    "no-nested-ternary": 0,
+    "no-multi-spaces": 0
```
Arcanemagus: I'm pretty strongly against disabling this rule... do you have a good case to do so?

.eslintrc.json
```diff
@@ -0,0 +1,35 @@
...
+    "max-len": [1, 150],
+    "no-console": 0,
+    "react/prefer-stateless-function": 0,
+    "react/prop-types": 0,
+    "react/jsx-no-bind": 0,
+    "react/jsx-closing-bracket-location": 0,
+    "react/no-deprecated": 0,
+    "react/jsx-boolean-value": 0,
+    "react/sort-comp": 0,
+    "prefer-template": 0,
```
Arcanemagus: Why is this being disabled?

.eslintrc.json
```diff
@@ -0,0 +1,35 @@
...
+  "extends": "airbnb",
+  "rules": {
+    "semi": [2, "never"],
+    "prefer-arrow-callback": 0,
+    "space-before-function-paren": [2, "never"],
+    "func-names": 0,
+    "no-param-reassign": 0,
+    "comma-dangle": [2, "never"],
+    "prefer-rest-params": 0,
+    "max-len": [1, 150],
```
Arcanemagus: Can this be changed to 120?

.eslintrc.json
```diff
@@ -0,0 +1,35 @@
...
+  "parser": "babel-eslint",
+  "extends": "airbnb",
+  "rules": {
+    "semi": [2, "never"],
+    "prefer-arrow-callback": 0,
+    "space-before-function-paren": [2, "never"],
+    "func-names": 0,
+    "no-param-reassign": 0,
+    "comma-dangle": [2, "never"],
+    "prefer-rest-params": 0,
```
Arcanemagus: Why is this being disabled?

.eslintrc.json
```diff
@@ -0,0 +1,35 @@
+{
+  "parser": "babel-eslint",
+  "extends": "airbnb",
+  "rules": {
+    "semi": [2, "never"],
+    "prefer-arrow-callback": 0,
+    "space-before-function-paren": [2, "never"],
+    "func-names": 0,
+    "no-param-reassign": 0,
```
Arcanemagus: Why in the world is this being disabled?

package.json
```diff
@@ -41,6 +42,12 @@
     "minimatch": "^3.0.0",
     "sb-debounce": "^1.0.1"
   },
+  "devDependencies": {
+    "babel-eslint": "^5.0.0",
```
Arcanemagus: If we want to take on a beta dependency, `babel-eslint@6.0.0-beta.6` supports `eslint@^2.4.0`.

.eslintrc.json
Arcanemagus: _Just_ realized this is `linter`, so it might actually be using React :stuck_out_tongue:.

.eslintrc.json
steelbrain: Should we create two `.eslintrc.json` files then?

.eslintrc.json
steelbrain: We need it here https://github.com/atom-community/linter/blob/8b71dddc444297d1b0eb5af8b2280bc98c7e46d9/lib/linter-registry.js#L43

.eslintrc.json
steelbrain: `messageKey` gets un-beautiful if we try to write it in templates https://github.com/atom-community/linter/blob/8b71dddc444297d1b0eb5af8b2280bc98c7e46d9/lib/helpers.js#L38

.eslintrc.json
steelbrain: Enabled

.eslintrc.json
steelbrain: Mainly to make optional parameters work, this is what we do now
```js
function a(b, c) {
  c = typeof c === 'string' ? c : 'something'
}
```
We'll have to do this instead
```js
function a(b, cGiven) {
  const c = typeof cGiven === 'string' ? cGiven : 'something'
}
```

I personally prefer the first one

package.json
steelbrain: Has it been released yet or still in beta?

steelbrain:
AppVeyor is failing because of some error in CI script for beta, Travis only failed on one release because of some mysterious reasons, they are both unrelated to this

.eslintrc.json
Arcanemagus: Exactly, `linter-eslint` is an example of this.

package.json
Arcanemagus: `babel-eslint` is being it's notoriously slow self and has yet to release this critical fix.

.eslintrc.json
Arcanemagus: The rule references [this article](http://eslint.org/docs/rules/no-param-reassign), which gives a great case as to why this is a bad idea. That being said, at least according to [this article](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments) (which you linked me :stuck_out_tongue:), if you don't _mention_ the `arguments` variable, and are in `"use strict"`, then you should be safe to disable this rule.

I personally prefer to avoid the potential issue, but as long as you are careful this should be fine to disable.

.eslintrc.json
Arcanemagus: I would just write that without the giant spacing, but I guess that's me :stuck_out_tongue:.

.eslintrc.json
Arcanemagus: Makes sense.

Arcanemagus:
Restarted that Travis build and it seems to have succeeded this time.

This looks good enough to me at this point, thanks!

Arcanemagus:
LGTM :ship:

PR #1100, opened 2016-03-18
Flowize linter

steelbrain:
Flowize the whole package

### DIFF
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -1,11 +1,17 @@
 'use babel'
 
+/* @flow */
+
 import { CompositeDisposable, Emitter } from 'atom'
+import type { Disposable } from 'atom'
 
 export default class Commands {
+  emitter: Emitter;
+  subscriptions: CompositeDisposable;
+
   constructor() {
-    this.subscriptions = new CompositeDisposable()
     this.emitter = new Emitter()
+    this.subscriptions = new CompositeDisposable()
 
     this.subscriptions.add(this.emitter)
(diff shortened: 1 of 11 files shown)


lib/commands.js
```diff
@@ -1,11 +1,17 @@
...
+import type { Disposable } from 'atom'
 
 export default class Commands {
+  emitter: Emitter;
+  subscriptions: CompositeDisposable;
+
   constructor() {
-    this.subscriptions = new CompositeDisposable()
     this.emitter = new Emitter()
+    this.subscriptions = new CompositeDisposable()
```
Arcanemagus: Not that there is really a _problem_ with this, but why re-order these?

Arcanemagus:
LGTM

lib/commands.js
steelbrain: I am moving the shorter variables above the longer ones so they look nice

joefitzgerald:
:-1: on this. What was the benefit? When was it decided to use flow on the package?

steelbrain:
@joefitzgerald The main benefit is that flow is smart, so if we have a class like this
```js
class Registry {
  add(something: Object)
}
```
and try to do
```js
registry.aa(someObj);
```
or
```js
registry.add('something')
```
it will error and tell us that we're doing it wrong.

It has been extremely helpful and it's autocomplete has been very accurate. We catch bugs everytime we move a package to flow. Can you state the reasons you are against this change?

PR #1101, opened 2016-03-21
Add severity attribute

### LINKED ISSUE #878: status line display can be misleading
When the `dartlang` plugin creates lints, it can (optionally) create 'info' level lints, as well as errors and warnings. These are things like todo comments, or suggestions to correct for best practices. We map this to the lint `Trace` type. If only info lints exist on a project, the status line UI makes the lints appear more series than they are:

screen shot 2015-09-07 at 10 35 05 am

If there are no issues at all, the status line shows as green w/ a check mark and the text `No Issues`. W/ info level lints, it shows as red, with an X, and the text `N Issues`.

It would be great to either give users a way to turn off the status line contribution (like the error panel), and/or to only have the status line contribution turn red iff there are errors or warnings.

steelbrain:
It has limited set of values, these are case sensitive, `error`, `warning` and `info`

They are not shown to the user but used for other internal purposes, we might even use these to paint the messages so the providers don't have to provide style for each message type.

Another benefit of having this attribute is that we can show in the UI how many are warnings and how many are infos

- [x] Add validation for the new attribute
- [x] Make message registry set the default value to `error`
- [x] Update wiki to include the new attribute

Fixes #878

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -58,6 +58,7 @@ export function fillMessage(message: Linter$Message, linterName: ?string) {
   message.name = message.name || linterName || null
   message.key = messageKey(message)
   if (!message.severity) {
-    message.severity = 'error'
+    message.severity = message.type === 'Warning' || message.type === 'warning' ? 'warning' :
+                       message.type === 'Info' || message.type === 'info' ? 'info' : 'error'
   }
 }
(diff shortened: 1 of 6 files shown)


Arcanemagus:
I'd always wondered why the message types weren't standardized, relying on them to match a predefined value for styling always seemed error prone to me. :+1:

Issue #1099, opened 2016-03-18
Project scoped linters have lazy updates


steelbrain:
File scoped linters work just fine but for project scoped ones, we have to save twice. Not sure if this is a UI bug or a linter bug.

This only happens on un-released master

steelbrain:
It looks like an issue with the debouncer, still no solid way to repro

steelbrain:
I have strong feeling this is fixed in https://github.com/steelbrain/linter-ui-default/pull/72
I haven't been able to repro it yet

steelbrain:
Turns out this is a bug in flow and hh, it has had a race condition for the last few months

PR #1121, opened 2016-04-03
Fix a race condition in using linter results

steelbrain:
The description of the bug is in the commit message fixing it

- [x] Fix the bug
- [x] Add specs for change
- [x] ~~Document change in changelog~~ (Not applicable since this bug was never released)


steelbrain:
This fix might also solve #1099 but not yet sure

PR #1126, opened 2016-04-10
Fix a bug in message registry

steelbrain:
Really hoping it will fix a long standing bug https://github.com/steelbrain/linter/issues/1099

This is a bug, specs confirm it, but not sure if it's exactly that bug


Arcanemagus:
LGTM

PR #1179, opened 2016-06-11
Search all panes for file to open; consolidate duplicated code

inkytonik:
Files are currently opened in the active pane, even if they are already open in another pane. To my mind it's better to reuse an open file even in another pane, since most people won't want to have the file open more than once. 

An extension could be to add an option to allow the "current pane" or "all panes" behaviour to be chosen on a per-user basis. I can add that if you'd like but I suspect it's not really needed.

The commit also reduces some obvious code duplication by introducing a gotoError method.

### DIFF
--- a/lib/commands.coffee
+++ b/lib/commands.coffee
@@ -68,27 +68,26 @@ class Commands
     # That way the index can be ++ an -- without caring about the array bounds.
     messages[index %% messages.length]
 
+  gotoError: (index) ->
+    message = @getMessage(@index)
+    return unless message?.filePath
+    return unless message?.range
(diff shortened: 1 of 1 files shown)


inkytonik:
I've looked at the failures. The build fails due to a test problem that existed on master before my commit, so I think that should be treated independently. 

Do you require a new test for my change?

steelbrain:
Yeah the failures in CI are unrelated to this PR. But it would still nice to have some tests for this one

inkytonik:
Ok, no problem. I'll look into some tests when I have a bit more time.

PR #1185, opened 2016-06-16
Refresh the tooltip when its current message gets removed

jviide:
`src/linter-view.js`'s `renderBubble` skips refreshing the tooltip when the cursor is within the currently shown tooltip's range. However the tooltip should get fully refreshed without skipping anything when its current message has been removed.

Here is an example of a situation when a stale tooltip is shown (using the current master 9338d172) even when the error it's showing gets fixed:

before

After the changes in this pull request the same situation looks like this:

after

The pull request tries to simplify internal state handling a bit (YMMV) by keeping all info related to the current tooltip in a single property called `bubble`. The property should be `null` when the tooltip is not shown. When the tooltip is shown then `bubble` should be an object containing all the relevant info (currently shown message, range & marker).

### DIFF
--- a/lib/linter-views.js
+++ b/lib/linter-views.js
@@ -16,7 +16,6 @@ export default class LinterViews {
     this.editors = editorRegistry
     this.bottomBar = null // To be added when status-bar service is consumed
     this.bubble = null
-    this.bubbleRange = null
 
     this.subscriptions.add(this.bottomPanel)
     this.subscriptions.add(this.bottomContainer)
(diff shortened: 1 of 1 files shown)


steelbrain:
What a coincidence, I just fixed this in the new linter UI like a few hours ago and was gonna backport the fix!

Anyways, failing CI is unrelated to this change. Thanks 👍

jviide:
Nice! Thanks 👍

steelbrain:
Released in v1.11.11 🎉

PR #1192, opened 2016-06-27
Updated all dependencies to most recent version, conformed eslint styles to new styleguide

amilajack:


.travis.yml
```diff
@@ -29,7 +29,3 @@ addons:
     - git
     - libgnome-keyring-dev
     - fakeroot
-
```
steelbrain: Any particular reason for removing this? This creates two CI builds for PRs open from branches of this repo

steelbrain:
Can you explain the rename from `.eslintrc.json` to `.eslintrc`?

amilajack:
Removing the prefix removes this issue:

```bash
Cannot read config file: /Users/amila/Documents/Projects/linter/.eslintrc.json
Error: Unexpected token i in JSON at position 250
```

It allows for a more flexible configuration without conforming to the json standard

steelbrain:
Funny thing I don't get that locally. Anyways, is there any particular benefit of merging the changes in this PR? As far as I can see it adds a lot of workarounds that are not necessary, maybe it's your local setup but they are not really necessary as they've been passing in the CI

amilajack:
Maybe it depends on your node version? Not sure

steelbrain:
It doesn't really depend on the node version, if I had to guess, you ran `npm install` on master and then started editing this branch, you should do a `rm -rf node_modules; npm install --development` and the errors you fixed here won't appear in the first place

.travis.yml
Arcanemagus: Very much a 👎 from me on this as well.

.eslintrc
```diff
@@ -2,8 +2,12 @@
   "parser": "babel-eslint",
   "extends": "airbnb/base",
   "rules": {
+    "no-underscore-dangle": 0,
+    "no-duplicate-imports": 0,
+    "global-require": 0,
     "semi": [2, "never"],
     "prefer-arrow-callback": 0,
+    "import/no-unresolved": [2, { ignore: ["atom"] }],
```
Arcanemagus: `npm` (and thus _any_ future Greenkeeper PR) will expand this out to multiple lines, run a `npm --save <package that is already listed>` to see what it will do.

steelbrain:
Closing based on discussion above, Thank you

PR #1206, opened 2016-07-20
Add apply function to solution

yitzchak:
Add ability to have custom apply function in v2 message solutions.

Supports steelbrain/linter-ui-default#113

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -22,10 +22,13 @@ export type Message = {
   reference: ?string,
   solutions?: Array<{
     title?: string,
-    position?: Range,
+    position: Range,
     currentText?: string,
-    replaceWith?: string,
-    apply: (() => void)
+    replaceWith: string,
+  } | {
+    title?: string,
+    position: Range,
+    apply: (() => any),
   }>,
   description?: Array<string>,
 }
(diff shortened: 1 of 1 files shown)


lib/types.js
```diff
@@ -21,10 +21,11 @@ export type Message = {
...
   solutions?: Array<{
-    title: ?string,
-    position: Range,
-    currentText: ?string,
-    replaceWith: string,
+    title?: string,
+    position?: Range,
+    currentText?: string,
+    replaceWith?: string,
+    apply: (() => void)
```
Arcanemagus: You're marking the callback as _required_, and changing the `replaceWith` to optional, if anything they should both be optional, with a note in the documentation that one or the other (but not both?) is required.

steelbrain:
Thanks @yitzchak

But the current type is not the ideal, ie. it makes the `position` optional and makes the `apply` callback required. Here's what you should change this too

```js
  solutions?: Array<{
    title?: string,
    position: Range,
    currentText?: string,
    replaceWith?: string,
  } | {
    title?: string,
    position: Range,
    apply: (() => any),
  }>,
```

steelbrain:
LGTM Thanks 👍

PR #1223, opened 2016-08-28
Transition towards an IDE state

steelbrain:
The branch is finally stable, it has full specs for everything in this package


steelbrain:
Will write docs in the morning, they'll be in the repo, wiki docs will be removed and would point to those docs instead.
Reference: https://github.com/steelbrain/linter/issues/1130

MikeNGarrett:
🤘

PR #1232, opened 2016-09-01
More validation in IndieDeligate

steelbrain:
Make sure supplied filePath and location.file is the same


steelbrain:
Github seems to be stuck with TravisCI even though it's hook has been removed and it's been disabled from Travis config. Merging in based on CircleCI pass

PR #1233, opened 2016-09-01
Rename lintOnFly to lintsOnChange for API v2

steelbrain:
Just like the title says


Arcanemagus:
Missed documenting the change in the upgrade guide text (but it _is_ in the example).

steelbrain:
Yeah that's what I thought, maybe I should still document it

PR #1244, opened 2016-09-14
Re-allow null from providers

### LINKED ISSUE #1243: Re-allow null returns from providers
In Linter API v1 a `null` or `undefined` value returned from a provider's `lint()` method would cause Linter to simply not update the results. This was determined to be a bug and was fixed for API v2, but it turns out there is a very valid use case for it.

Please modify API v2 to support a return of `null` signifying the messages for a provider shouldn't be updated. Thanks!

steelbrain:
Fixes #1243

### DIFF
--- a/docs/types/standard-linter-v2.md
+++ b/docs/types/standard-linter-v2.md
@@ -70,4 +70,4 @@ regardless of the scope of the current file.
 
 **Q**: How do I make linter use last messages for my provider?
 
-**A**: Return `null` from your provider, returning `false` or `undefined` will result in an error.
+**A**: Return `null` from your provider to make it re-use last messages.
--- a/docs/types/standard-linter-v2.md
+++ b/docs/types/standard-linter-v2.md
@@ -67,3 +67,7 @@ array.
 
 **A**: Use `['*']` as the `grammarScopes` value, and it'll be triggered
 regardless of the scope of the current file.
+
+**Q**: How do I make linter use last messages for my provider?
+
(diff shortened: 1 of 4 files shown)


docs/types/standard-linter-v2.md
```diff
@@ -67,3 +67,7 @@ array.
 
 **A**: Use `['*']` as the `grammarScopes` value, and it'll be triggered
 regardless of the scope of the current file.
+
+**Q**: How do I make linter use last messages for my provider?
+
+**A**: Return `null` from your provider, returning `false` or `undefined` will result in an error.
```
Arcanemagus: I'd cut out the part about `false` or `undefined`, this page is just meant to describe the right way to do things, no point in mentioning the wrong way 😛.

spec/linter-registry-spec.js
```diff
@@ -262,14 +263,97 @@ describe('LinterRegistry', function() {
...
+      expect(timesFinished).toBe(1)
+      promise = linterRegistry.lint({ editor, onChange: false })
+      expect(timesBegan).toBe(2)
+      expect(timesUpdated).toBe(1)
+      expect(timesFinished).toBe(1)
+      expect(await promise).toBe(true)
+      expect(timesUpdated).toBe(1)
+      expect(timesFinished).toBe(2)
+    })
+    it('cries if response is not array and is non-null', async function() {
```
Arcanemagus: Maybe `'cries if response is not an Array or null'`? The current wording works, but is a bit awkward.

PR #1258, opened 2016-10-17
Fix an inconsistent behavior bug with on change providers

steelbrain:
They would be triggered on editor open without even changing anything. This fixes that behavior

I tried to add a spec, but wasn't able to repro the bug in spec window. I suppose it happens in the first place because of some internal atom bug or something. In any case, I've added a note above the fix to make sure it's not deleted in some refactoring


Arcanemagus:
Interesting, previously it was theorized that it was `editor-config` that was causing this, but I don't think anyone actually verified it.

PR #1269, opened 2016-11-18
Add documentation for the v1 API to the repo

Arcanemagus:
**NEEDS WORK**

A first pass at adding documentation for the v1 API to the repo.

The Indie API most likely needs some work as I am not as familiar with that aspect.

Note that once this is merged the v2 documentation can be updated to point here instead of the Wiki, and the Wiki can be completely deprecated.


steelbrain:
Hi @Arcanemagus. Thank you so much for working on this, I've pushed in some changes. Let me know what you think

Arcanemagus:
LGTM!

PR #1287, opened 2016-12-21
Add priority attribute to solutions

steelbrain:
Laying foundations for https://github.com/steelbrain/linter-ui-default/issues/131


steelbrain:
This should be a non-controversial change so merging in

PR #1289, opened 2016-12-22
Minor changes to API and docs

steelbrain:
As discussed with @Arcanemagus

### DIFF
--- a/docs/types/linter-message-v2.md
+++ b/docs/types/linter-message-v2.md
@@ -23,7 +23,7 @@ type Message = {
     file: string,
     position?: Point,
   },
-  // ^ Reference link to the issue, useful for jumping to classes etc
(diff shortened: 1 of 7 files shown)


docs/types/linter-message-v2.md
```diff
@@ -7,24 +7,31 @@ This document describes the type of Linter Message v2. It's supported in
...
     file: string,
     position: Range,
   },
-  source?: {
+  // ^ Location of the issue (aka where to highlight)
+  reference?: {
     file: string,
     position?: Point,
   },
+  // ^ Reference link to the issue, useful for jumping to classes etc
```
Arcanemagus: How about this:
"Reference to a different location in the editor, useful for jumping to classes etc."

PR #1319, opened 2017-02-12
Use contrast as text color

### LINKED ISSUE #1317: Possible Coloring Issue?
Hi @steelbrain 
I just noticed the text in warning tooltips in linter-flake8 ([Issue](https://github.com/AtomLinter/linter-flake8/issues/347#issuecomment-279191496)) is really hard to see when using `material-ui` theme.
Just wonder if there is a reason why you forced the text to be white in [here](https://github.com/steelbrain/linter/blob/v1.11.21/styles/linter-plus.less#L165)?

image

steelbrain:
Fixes #1317

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.11.22
+
+- Use contrast colors in inline tooltip instead of hard-coded white (#1317)
+
 ## 1.11.21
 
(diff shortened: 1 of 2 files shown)


Arcanemagus:
Adding a few screenshots from our private chat for examples:
image
image
image

PR #1326, opened 2017-02-21
Upgrade to latest package-deps

steelbrain:
Also remove a now unnecessary config

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,8 +9,9 @@ export default {
   instance: null,
 
   activate(state: State) {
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -9,8 +9,9 @@ export default {
   instance: null,
 
   activate(state: State) {
-    if (!atom.inSpecMode() && atom.config.get('linter.installRecommendedPackages')) {
-      require('atom-package-deps').install('linter') // eslint-disable-line
+    if (!atom.inSpecMode()) {
+      // eslint-disable-next-line
```
Arcanemagus: Disable the specific rule that is causing issues, not _all_ checking.

PR #1331, opened 2017-03-01
Add command to show registered providers for current file type

### LINKED ISSUE #1329: Add command to show linters of current file type as notification
Just like the title says

steelbrain:
Fixes #1329

### DIFF
--- a/lib/linter-registry.js
+++ b/lib/linter-registry.js
@@ -3,7 +3,7 @@
 import { Emitter, CompositeDisposable } from 'atom'
 import type { TextEditor, Disposable } from 'atom'
 import { linter as validateLinter, messages as validateMessages, messagesLegacy as validateMessagesLegacy } from './validate'
-import { normalizeMessages, normalizeMessagesLegacy, isPathIgnored, shouldTriggerLinter, $version, $activated, $requestLatest, $requestLastReceived } from './helpers'
+import { getEditorScopes, normalizeMessages, normalizeMessagesLegacy, isPathIgnored, shouldTriggerLinter, $version, $activated, $requestLatest, $requestLastReceived } from './helpers'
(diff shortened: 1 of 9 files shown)


lib/linter-registry.js
```diff
@@ -71,8 +74,7 @@ export default class LinterRegistry {
       return false
     }
 
-    const scopes = editor.scopeDescriptorForBufferPosition(editor.getCursorBufferPosition()).getScopesArray()
-    scopes.push('*')
+    const scopes = editor.scopeDescriptorForBufferPosition(editor.getCursorBufferPosition()).getScopesArray().concat(['*'])
```
Arcanemagus: What about using this to support all cursors instead of just the last one?

```js
const scopes = editor.getCursors().reduce((scopes, cursor) => (
    scopes.concat(cursor.getScopeDescriptor().getScopesArray())
  ), []).concat(['*']);
```

_Note: This list isn't unique, there will likely be duplicates if there is more than one active cursor._

lib/main.js
```diff
@@ -53,6 +54,13 @@ class Linter {
         this.registryEditors.createFromTextEditor(textEditor)
       }
     })
+    this.commands.onShouldShowLinterProviders(() => {
+      const linters = this.registryLinters.getLinters()
+      const textEditor = atom.workspace.getActiveTextEditor()
+      const textEditorScopes = textEditor.scopeDescriptorForBufferPosition(textEditor.getCursorBufferPosition()).getScopesArray().concat(['*'])
```
Arcanemagus: Should this be moved to a helper function? This code is used in more than one place.

steelbrain:
The CI job associated has completed successfully, GitHub's webhooks must be having some issues so merging in

PR #1335, opened 2017-03-03
Use the new pending state way

### LINKED ISSUE #1334: Lint Preview Tabs stops all linting
It seems that the "Lint Preview Tabs" setting is broken, if it is disabled then _all_ linting is stopped, not just on preview tabs.

steelbrain:
Fixes #1334

### DIFF
--- a/lib/linter-registry.js
+++ b/lib/linter-registry.js
@@ -66,10 +66,10 @@ export default class LinterRegistry {
     const filePath = editor.getPath()
 
     if (
-      (onChange && !this.lintOnChange) ||                                          // Lint-on-change mismatch
-      !filePath ||                                                                 // Not saved anywhere yet
-      isPathIgnored(editor.getPath(), this.ignoreGlob, this.ignoreVCS) ||          // Ignored by VCS or Glob
-      (!this.lintPreviewTabs && editor.hasTerminatedPendingState === false)        // Ignore Preview tabs
+      (onChange && !this.lintOnChange) ||                                                       // Lint-on-change mismatch
+      !filePath ||                                                                              // Not saved anywhere yet
+      isPathIgnored(editor.getPath(), this.ignoreGlob, this.ignoreVCS) ||                       // Ignored by VCS or Glob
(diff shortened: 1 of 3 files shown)


lib/linter-registry.js
```diff
@@ -66,10 +66,10 @@ export default class LinterRegistry {
...
 
     if (
-      (onChange && !this.lintOnChange) ||                                          // Lint-on-change mismatch
-      !filePath ||                                                                 // Not saved anywhere yet
-      isPathIgnored(editor.getPath(), this.ignoreGlob, this.ignoreVCS) ||          // Ignored by VCS or Glob
-      (!this.lintPreviewTabs && editor.hasTerminatedPendingState === false)        // Ignore Preview tabs
+      (onChange && !this.lintOnChange) ||                                                       // Lint-on-change mismatch
+      !filePath ||                                                                              // Not saved anywhere yet
+      isPathIgnored(editor.getPath(), this.ignoreGlob, this.ignoreVCS) ||                       // Ignored by VCS or Glob
+      (!this.lintPreviewTabs && atom.workspace.getActivePane().getPendingItem() === editor)     // Ignore Preview tabs
```
Arcanemagus: Is this going to work in the older style that still has a working `editor.hasTerminatedPendingState`?

steelbrain: We've bumped our minimum required atom version quite high

Arcanemagus: Might be best to follow the logic `tab-view` [uses here](https://github.com/atom/tabs/blob/4c94444616f0ea6f851f62fd825c9c114669e988/lib/tab-view.coffee#L231).

Arcanemagus: Looks like the pending item stuff was added in 1.6.0, with some refactoring in 1.7.0. So our minimum version of v1.4.0 doesn't even know pending items _exist_ 😛.

steelbrain: Solved by increasing minimum version 😛

steelbrain:
Merging in as the primary concern has been resolved

PR #1341, opened 2017-03-06
Add commands to enable/disable specific linters

### LINKED ISSUE #1165: Add Option to Enable/Disable a Specific Linter
It would be immensely to be able to toggle specific linters manually when dealing with JSLint/JSHint/ESLint/JSCS across different projects. In my opinion, SublimeLinter has a good reference implementation:

Select enable or disable from the menu:

screen shot 2016-05-30 at 9 14 34 am

Select linter to disable:

screen shot 2016-05-30 at 9 19 16 am

steelbrain:
Fixes #1165

- [x] Implement view
- [x] Respect it in linter registry
- [x] Add specs


lib/commands.js
```diff
@@ -2,34 +2,41 @@
...
 
-  constructor(state: State) {
-    this.state = state
+  constructor() {
     this.emitter = new Emitter()
     this.subscriptions = new CompositeDisposable()
 
     this.subscriptions.add(this.emitter)
+    this.subscriptions.add(atom.commands.add('atom-workspace', {
+      'linter:debug': () => this.debug(),
```
Arcanemagus: `debug()` needs a `TextEditor` instance, which it could be triggered without one from here, is there a reason it was moved?

.eslintrc.json
```diff
@@ -5,6 +5,7 @@
     "no-param-reassign": 0,
     "no-duplicate-imports": 0,
     "no-underscore-dangle": 0,
+    "import/no-duplicates": 0,
```
Arcanemagus: Unless this is being used _everywhere_, this should be done as a comment where it is necessary.

steelbrain:
Here's how it looks like in action, had to reduce gif quality because of max filesize limits on GitHub

Linter Toggle in action

lib/toggle-view.js
```diff
@@ -0,0 +1,90 @@
...
+        prettyPrint: true,
+        createIfNonExistent: false,
+      })
+    }
+    return this.config
+  }
+  async getItems(): Promise<Array<string>> {
+    const disabled = await (await this.getConfig()).get('disabled')
+    if (this.action === 'disable') {
+      return this.providers.filter(name => disabled.indexOf(name) === -1)
```
Arcanemagus: You could just do `disabled.includes(name)` instead of the older `disabled.indexOf(name) === -1`, leads to clearer code.

spec/linter-registry-spec.js
```diff
@@ -488,12 +503,68 @@ describe('LinterRegistry', function() {
...
       const promise = linterRegistry.lint({ editor, onChange: false })
+      await wait(20)
       expect(timesBegan).toBe(1)
       expect(timesUpdated).toBe(0)
       expect(timesFinished).toBe(0)
       expect(await promise).toBe(true)
       expect(timesUpdated).toBe(1)
       expect(timesFinished).toBe(1)
     })
+    it('does not invoke a linter if its ignored', async function() {
```
Arcanemagus: "its" -> "it's"

lib/commands.js
steelbrain: Good catch!

lib/linter-registry.js
```diff
@@ -1,12 +1,17 @@
...
 
+import ConfigFile from 'sb-config-file'
 import { Emitter, CompositeDisposable } from 'atom'
 import type { TextEditor, Disposable } from 'atom'
-import { linter as validateLinter, messages as validateMessages, messagesLegacy as validateMessagesLegacy } from './validate'
-import { getEditorCursorScopes, normalizeMessages, normalizeMessagesLegacy, isPathIgnored, shouldTriggerLinter, $version, $activated, $requestLatest, $requestLastReceived } from './helpers'
+
+import * as Helpers from './helpers'
+import * as Validate from './validate'
+import { $version, $activated, $requestLatest, $requestLastReceived, LINTER_CONFIG_FILE_PATH, LINTER_CONFIG_FILE_DEFAULT } from './helpers'
```
Arcanemagus: If you put this above this line, it documents _exactly_ why you need to disable the rule:

```raw
// eslint-disable-next-line import/no-duplicates
```

steelbrain: This particular type of error is actually annoying to fix with `eslint-disable-next-line` comment, because you have to do it twice on both imports instead of once so I just disabled it in this file entirely

Arcanemagus: Lovely...

CoWinkKeyDinkInc:
I don't see this option when using `command-shift P`.  I only have linter-write-good installed so I think it might be because I don't have another linter plugin installed, or is because update not pushed into Atom yet?

steelbrain:
Hi @CoWinkKeyDinkInc

It's most likely because you are on the APM version of linter, We're going to push out Linter v2 in less than 24 hours so you can then upgrade your linter to have these commands

PR #1346, opened 2017-03-13
Add Greeter

steelbrain:
Also add a v2 message for it

### DIFF
--- a/lib/greeter/greet-v2-welcome.js
+++ b/lib/greeter/greet-v2-welcome.js
@@ -1,0 +1,17 @@
+/* @flow */
+
+import coolTrim from 'cool-trim'
+
+export default function greet() {
+  return atom.notifications.addInfo('Welcome to Linter v2', {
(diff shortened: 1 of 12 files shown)


lib/greeter/greet-v2-welcome.js
```diff
@@ -0,0 +1,20 @@
...
+import coolTrim from 'cool-trim'
+
+export default function greet() {
+  return atom.notifications.addInfo('Welcome to Linter v2', {
+    dismissable: true,
+    description: coolTrim`
+      Hi Linter user! 👋
+
+      Linter has been upgraded to v2.
+      TL;DR is that UI has been split into a separate package and the base package has been rewritten to maximize stability.
```
Arcanemagus: Do users really need to care about this? For them the entire idea is that it "just works".

steelbrain: Taken care of

PR #1347, opened 2017-03-14
Prepare docs for v2

steelbrain:
Just like the title says


steelbrain:
Merging on LGTM from @Arcanemagus

PR #1348, opened 2017-03-14
Upgrade ESLint config

steelbrain:
I feel bad for having a PR this small, but couldn't slip this in in other PRs because PRs should be focused


steelbrain:
Merging on LGTM from @Arcanemagus

PR #1353, opened 2017-03-14
Add support for legacy Indie v1 providers

steelbrain:
A day after the v2 release, better late than never right?!

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -56,7 +56,12 @@ export default {
   },
   provideIndie(): Object {
     return indie =>
-      instance.registryIndie.register(indie)
+      instance.registryIndie.register(indie, 2)
+  },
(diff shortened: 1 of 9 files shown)


lib/index.js
```diff
@@ -5,27 +5,28 @@ import Linter from './main'
 import Greeter from './greeter'
 import type { UI, Linter as LinterProvider } from './types'
 
+let greeter
```
steelbrain: This change was required to get flow to report errors, it was previously ignoring all calls to `this.greeter` and `this.instance`, now it reports correctly

steelbrain:
Self reviewed multiple times and looks good, need it in for future work so merging in, it has specs for every new change and older ones too so shouldn't be controversial

PR #1371, opened 2017-03-16
Make name optional for v1 again

steelbrain:
Fixes https://github.com/atom-haskell/haskell-ghc-mod/issues/208


Arcanemagus:
Looks like the specs need some attention from this change, the code itself LGTM though.

PR #1391, opened 2017-03-22
Only show welcome if older configs are present

### LINKED ISSUE #1377: Use Atom builtin config.set/get instead of adding another arbitrary JSON file to track
Why create `ATOM_HOME/linter-config.json` when you can [should] just use `atom.config.set`/`get`?

Same reasoning as https://github.com/steelbrain/package-deps/pull/89

### LINKED ISSUE #1381: Only show upgrade message if old settings detected
Only show the upgrade message if the old UI related settings are found, if those around found assume it is a _new_ user and don't spam them with something they have no idea about anyway 😛.

steelbrain:
Fixes #1381
Fixes #1377

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,6 +10,9 @@ let instance
 
 export default {
   activate() {
+    greeter = new Greeter()
+
+    const linterConfigs = atom.config.get('linter')
     // Unset v1 configs
     const removedV1Configs = [
       'lintOnFly',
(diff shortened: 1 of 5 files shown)


lib/index.js
```diff
@@ -10,6 +10,9 @@ let instance
 
 export default {
   activate() {
+    greeter = new Greeter()
```
Arcanemagus: This is only used down on L41, why not just define it as a `const` in that `if`?

steelbrain: Because it has to be disposed on `deactivate`, see below

Arcanemagus: Ha, whoops!

PR #1414, opened 2017-04-03
Add instructions for open dev tools in error message

### LINKED ISSUE #1412: Enhance the error message
When a provider doesn't catch an error, the current message just tells users to "See console for more info". As we have users that may not be familiar with web development this should be clarified to "Open View -> Developer - Toggle Developer Tools and check the Console tab for more details", or something similar to point them in the right direction.

steelbrain:
Fixes #1412

### DIFF
--- a/lib/linter-registry.js
+++ b/lib/linter-registry.js
@@ -139,7 +139,7 @@ export default class LinterRegistry {
       }, (error) => {
         this.emitter.emit('did-finish-linting', { number, linter, filePath: statusFilePath })
         atom.notifications.addError(`[Linter] Error running ${linter.name}`, {
-          detail: 'See console for more info',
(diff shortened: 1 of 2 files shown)


lib/linter-registry.js
```diff
@@ -139,7 +139,7 @@ export default class LinterRegistry {
       }, (error) => {
         this.emitter.emit('did-finish-linting', { number, linter, filePath: statusFilePath })
         atom.notifications.addError(`[Linter] Error running ${linter.name}`, {
-          detail: 'See console for more info',
+          detail: 'See console for more info. (Open View -> Developer -> Toogle Developer Tools)',
```
Arcanemagus: "console" -> "Console".

PR #1422, opened 2017-04-10
Remove linter state file

steelbrain:
Fixes https://github.com/steelbrain/linter/issues/1377

Also merges in contents of the file if it exists, before removing it


package.json
```diff
@@ -57,6 +57,16 @@
...
   "configSchema": {
+    "disabledProviders": {
+      "title": "Disabled providers",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "description": "Names of disabled linter providers",
+      "default": [],
+      "order": 1
```
Arcanemagus: Is there any reason this is at the top and not the bottom?

steelbrain: Moved to the bottom

PR #1463, opened 2017-05-11
chore(package): update jasmine-fix to version 1.1.0

steelbrain:


steelbrain:
Will update to v1.1.1 instead

PR #1493, opened 2017-06-16
Allow messages to override the linterName (in v2)

### LINKED ISSUE #1485: Allow overriding `linterName` in messages?
In the Language Service Protocol (https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md), diagnostics (`interface Diagnostic`) can specify dynamic sources.

One use case for this might be for a JavaScript language server, where diagnostics can come from both a typing system (Flow, TypeScript) and linters (ESLint); it makes sense to distinguish between them.

When bridging this with the Linter API (as https://github.com/atom/atom-languageclient is doing), this presents a slight issue as it's designed around registering fixed source names. It's certainly possible to dynamically register linters as diagnostics come in (especially with Indie v2), but that's rather heavyweight.

Would it be reasonable to allow for users to override `linterName` in the Message v2 API (from https://github.com/steelbrain/linter/blob/master/docs/types/linter-message-v2.md)? The semantics would be pretty straightforward - if you provide a custom `linterName` then Linter would use it, but otherwise it'd default to the name of the linter.

See: https://github.com/atom/atom-languageclient/issues/47

hansonw:
This PR allows providers to override the displayed `linterName` on a per-message level.
I added a spec as well as a validator+spec.

Released under CC0.
Closes #1485

### DIFF
--- a/docs/types/linter-message-v2.md
+++ b/docs/types/linter-message-v2.md
@@ -44,12 +44,13 @@ type Message = {
   description?: string | (() => Promise<string> | string)
   // ^ Markdown long description of the error, accepts callback so you can do
   // http requests etc.
+  linterName?: string,
+  // ^ Optionally override the displayed linter name. Defaults to provider name.
 
   // NOTE: DO NOT SPECIFY THESE IN PROVIDER
   // Automatically added by base linter for UI consumers
   key: string,
   version: 2,
-  linterName: string,
 }
 ```
 
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -98,7 +98,9 @@ export function normalizeMessages(linterName: string, messages: Array<Message>) 
       }
     }
     message.version = 2
-    message.linterName = linterName
+    if (!message.linterName) {
+      message.linterName = linterName
+    }
     message.key = messageKey(message)
   }
 }
--- a/lib/validate/index.js
+++ b/lib/validate/index.js
@@ -103,6 +103,7 @@ function validateMessages(linterName: string, entries: Array<Message>): boolean 
     let invalidSolution = false
     let invalidReference = false
     let invalidDescription = false
+    let invalidLinterName = false
 
     for (let i = 0, length = entries.length; i < length; ++i) {
       const message = entries[i]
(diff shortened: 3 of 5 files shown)


docs/types/linter-message-v2.md
```diff
@@ -44,12 +44,13 @@ type Message = {
   description?: string | (() => Promise<string> | string)
   // ^ Markdown long description of the error, accepts callback so you can do
   // http requests etc.
+  linterName?: string,
+  // ^ Optionally override the displayed linter name. (Defaults to provider)
```
Arcanemagus: "Defaults to provider name"

docs/types/linter-message-v2.md
```diff
@@ -44,12 +44,13 @@ type Message = {
   description?: string | (() => Promise<string> | string)
   // ^ Markdown long description of the error, accepts callback so you can do
   // http requests etc.
+  linterName?: string,
+  // ^ Optionally override the displayed linter name. Defaults to provider name
```
Arcanemagus: Add a `.` on the end, or bring back the `()`, either way works.

Arcanemagus:
Looks like the AppVeyor failures are due to the new async nature of `TextEditor::save()`, putting up a PR fixing the spec for that.

hansonw:
:+1: done

Arcanemagus:
@steelbrain #1494 should be merged before this, or `master` builds will fail.

lib/helpers.js
```diff
@@ -98,7 +98,7 @@ export function normalizeMessages(linterName: string, messages: Array<Message>)
       }
     }
     message.version = 2
-    message.linterName = linterName
+    message.linterName = message.linterName || linterName
```
steelbrain: Can we instead do
```
if (!message.linterName) {
  message.linterName = linterName
}
```

Really minor but personal preference :)

hansonw:
Done!

steelbrain:
Perfect 👌 Now a rebase on master/merge from master and it'll be good to land 😃

steelbrain:
Thanks again for working on this 👍

steelbrain:
Released in v2.2.0 🎉

PR #1494, opened 2017-06-16
Make EditorLinter::onShouldLint spec async

Arcanemagus:
Instead of relying on timing and bad counting, make the spec truly async and properly count the number of invocations.


steelbrain:
Thank You

PR #1504, opened 2017-06-28
Include description in message key

nwolverson:
Fix issue where a lint message description changes but the excerpt, range, etc do not change and so the message does not update (https://github.com/nwolverson/atom-ide-purescript/issues/164).

(This also seems to occur if I call `clearMessages` prior to adding the updated message)


steelbrain:
Can you also include this in CHANGELOG under `Upcoming` title?

nwolverson:
Sure, updated.

PR #1525, opened 2017-08-16
[WIP] Project wide linting

steelbrain:
IT WORKS PEOPLE! IT'S WORKING

### Old stuff

- [ ] Add global and provider-provided `ignoreFileSources`
- [ ] Add `lintsProject` bool to providers
- [ ] Validate these two types on providers
- [ ] Watch all dirs in project for non-ignored files
- [ ] Lint them several at a time (configurable or number of core count?)
- [ ] Save their results and show them at boot, tho re-lint files in the background, this time less at a time
- [ ] Make this toggleable from config


steelbrain:
Works flawlessly so far, next up is trimming the results of things like fix callbacks and serializing them with Atom state so we resume with a nice state and only lint the files that have changed since the last time (or maybe re-lint all but with delays to keep cpu usage down?)

steelbrain:
Note to self: Use Atom watcher APIs instead of chokidar

steelbrain:
Looks like the Atom file watcher API hasn't been released yet so will have to keep using chokidar for now https://github.com/atom/atom/pull/14853

steelbrain:
No longer pursuing this.

PR #1532, opened 2017-08-21
Cleanup

steelbrain:
- Removes old config removal code
- Removes greeter


steelbrain:
Nothing controversial in this PR so gonna merge it in

PR #1551, opened 2017-10-03
Exclude non-essential files

vinkla:
This way the users of this package doesn't have to download non-essential files which aren't necessary for the package to function. Such as files used for development (docs, tests, .travis.yml, etc).

Happy Hacktoberfest!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,6 +12,4 @@
 /appveyor.yml export-ignore
 /circle.yml export-ignore
 /package-lock.json export-ignore
-/CHANGELOG.md export-ignore
 /CONTRIBUTING.md export-ignore
-/README.md export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,7 +6,6 @@
 /.editorconfig export-ignore
 /.eslintignore export-ignore
 /.eslintrc.json export-ignore
-/.eslintrc.json export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,2 +1,17 @@
...
+/.editorconfig export-ignore
+/.eslintignore export-ignore
+/.eslintrc.json export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.travis.yml export-ignore
+/appveyor.yml export-ignore
+/circle.yml export-ignore
+/package-lock.json export-ignore
+/CHANGELOG.md export-ignore
```
steelbrain: I would personally like to keep CHANGELOG and README in the export because users locally see it in settings view

vinkla: I've pushed an update!

steelbrain:
Thank You

.gitattributes
```diff
@@ -1,2 +1,15 @@
...
+/spec export-ignore
+/.editorconfig export-ignore
+/.eslintignore export-ignore
+/.eslintrc.json export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.travis.yml export-ignore
+/appveyor.yml export-ignore
+/circle.yml export-ignore
+/package-lock.json export-ignore
```
Arcanemagus: Although not used by the `npm@3` wrapped by `apm`, this should probably still be exported for when `apm` does update...

PR #1552, opened 2017-10-06
Lint on TextBuffer::onDidReload

### LINKED ISSUE #1550: Remove linter positions when file updates externally (using Lint on Save)
I'm currently using lint on save rather than on type as eslint is a bit slow, but something annoying with it is that if the file changes externally (e.g. `git pull`) Atom will update the file in the editor but the linter positions will continue to be left in position from the previous file state

There's two possible things that could be done either of which would be nicer at least for me:

1. Simply remove the linter positions and leave it as if it weren't linted at all until the next save
2. (My Preferred) Re-lint when file changes externally

Arcanemagus:
When the TextBuffer reloads the document from disk trigger a lint request automatically.

Fixes #1550.

### DIFF
--- a/lib/editor-linter.js
+++ b/lib/editor-linter.js
@@ -22,9 +22,12 @@ export default class EditorLinter {
     this.subscriptions.add(this.editor.onDidDestroy(() =>
       this.dispose(),
     ))
-    this.subscriptions.add(this.editor.onDidSave(debounce(() =>
-      this.emitter.emit('should-lint', false),
-    ), 16, true))
+    this.subscriptions.add(this.editor.onDidSave(
+      debounce(() => this.emitter.emit('should-lint', false), 16, true),
+    ))
+    this.subscriptions.add(this.editor.getBuffer().onDidReload(
+      debounce(() => this.emitter.emit('should-lint', false), 16, true),
+    ))
     // NOTE: TextEditor::onDidChange immediately invokes the callback if the text editor was *just* created
     // Using TextBuffer::onDidChange doesn't have the same behavior so using it instead.
     this.subscriptions.add(subscriptiveObserve(atom.config, 'linter.lintOnChangeInterval', interval =>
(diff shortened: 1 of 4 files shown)


lib/editor-linter.js
```diff
@@ -22,9 +22,12 @@ export default class EditorLinter {
...
       this.dispose(),
     ))
-    this.subscriptions.add(this.editor.onDidSave(debounce(() =>
-      this.emitter.emit('should-lint', false),
-    ), 16, true))
+    this.subscriptions.add(this.editor.onDidSave(
+      debounce(() => this.emitter.emit('should-lint', false), 16, true),
+    ))
+    this.subscriptions.add(this.editor.getBuffer().onDidReload(
+      debounce(() => this.emitter.emit('should-lint', false), 16, true),
```
steelbrain: Can we change the debounce parameters to `1000` and `false`, so it does not invoke until onDidReload stops invoking, and calls the callback 1 second after the last call

Arcanemagus:
This works locally and on AppVeyor, trying to determine what is going wrong on CircleCI the long way...

spec/editor-linter-spec.js
```diff
@@ -43,26 +96,50 @@ describe('EditorLinter', function() {
...
+
+      expect(timesTriggered).toBe(0)
+      modifyFile(tempPath)
+      await waitForShouldLint()
+      modifyFile(tempPath)
+      await waitForShouldLint()
+      expect(timesTriggered).toBe(2)
+
+      rimraf.sync(tempDir)
+    })
```
steelbrain: Can you add a third parameter options here and specify a timeout of 1.5 seconds?

Arcanemagus: I can try that, but when SSHing in and trying different debugging steps the test is timing out on the very first call to `waitForShouldLint`.

steelbrain: I guess trying doesn't hurt 😛

PR #1573, opened 2018-02-21
[Docs] Can't return legacy message from v2 provider

### LINKED ISSUE #1572: Documentation clarification: can you return legacy messages from standard V2 provider?
Looking at the [source code](https://github.com/steelbrain/linter/blob/7d7e52625da118f51df73b5083d751a047effd44/lib/linter-registry.js#L126) I'm convinced you can't and everything will break horribly if you try. But the [documentation](https://steelbrain.me/linter/types/standard-linter-v2.html) says `lint` can return `Array<Message | MessageLegacy>`.

So... either I'm misunderstanding the code, or documentation might need fixing.

lierdakil:
Closes #1572 

@Arcanemagus uh... kinda like this? I mean, it's a trivial change.

### DIFF
--- a/docs/types/standard-linter-v2.md
+++ b/docs/types/standard-linter-v2.md
@@ -21,9 +21,9 @@ type Linter = {
   lint(
     textEditor: TextEditor
   ):
-    | Array<Message | MessageLegacy>
+    | Array<Message>
     | null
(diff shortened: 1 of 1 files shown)


steelbrain:
Thanks @lierdakil :)

PR #1581, opened 2018-05-08
Add button to open Console

idleberg:
It's probably easier for a user to open the console by the click of the button, rather than providing a description where to find a toggle to do so.


steelbrain:
Thank you for working on this!

Issue #1510, opened 2017-07-13
Closing a TextEditor wipes all messages for that file


Arcanemagus:
When you close a `TextEditor` instance it seems that `linter` is removing all messages associated with that `TextEditor` which is great if that was the only instance of that file open.

If however there are multiple `TextEditor`s open to a single file (`TextBuffer`), then the messages are cleared on the other `TextEditor`s and need another `lint()` call to recreate them.

Steps to reproduce:
1. Open a file, create some linter messages in it
2. Split the file into multiple editors
3. Close the second one of them

Expected:
The first instance on the file still shows the same messages it did before the second instance was closed.

Actual:
The messages are cleared from the first instance, requiring `lint` to re-run to recreate them.

Seen with:
* `linter-ui-default`
* `linter-ui-plus`
* `atom-minimap-linter`

steelbrain:
This has been fixed in `master`

PR #1598, opened 2018-10-12
Don't wipe messages for file if it is opened multiple times

ericcornelissen:
This addresses the issue described in #1510 

## Preview
This first gif shows the current behaviour, the two JS editors are for the same file. Notice how the linting errors in the bottom now longer show up when focusing the editor at the top.
current behaviour

This second gif shows the new behaviour, the two JS editors are for the same file. Notice how the linting errors in the bottom still show up when focusing the editor at the top.
new bahaviour

## Alternative implementations
I'm not sure if the implementation I chose is desired. My reasoning behind it is that the information (`thereIsAnotherEditorForTheSameFile`) might be useful in other scenarios as well. Alternatively the check for other buffers can be performed in the [deleteByBuffer](https://github.com/steelbrain/linter/blob/1f9efea649c43ab50097d6ddc0d1e9685ee28413/lib/message-registry.js#L136) in the `MessageRegistry`.

Also, let me know if the variable name `thereIsAnotherEditorForTheSameFile` is too much 😛 

## Limitations ⚠️
This PR does not fix everything noted in or related to #1510. For example, in the second gif you can see that the lint-gutter of the editor at the top stops working when the editor at the bottom is closed.

### DIFF
--- a/lib/editor-linter.js
+++ b/lib/editor-linter.js
@@ -46,7 +46,14 @@ export default class EditorLinter {
     return this.emitter.on('did-destroy', callback)
   }
   dispose() {
-    this.emitter.emit('did-destroy')
+    let thereIsAnotherEditorForTheSameFile = false
+    for (const editor of atom.textEditors.editors) {
+      if (this.editor.getTitle() === editor.getTitle()) {
+        thereIsAnotherEditorForTheSameFile = true
+      }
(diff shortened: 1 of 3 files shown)


steelbrain:
Thank you for working on this!

PR #1682, opened 2020-02-18
Use repositoryForDirectory from atom.project

steelbrain:
This has been available in Atom API for years and lets us reduce some of the complexity in our helpers and be future-proof as Atom adds different directory providers


steelbrain:
`repositoryForDirectory` is an async API, which doesn't make sense in our case. Keeping existing code 🤦‍♂

PR #1706, opened 2020-08-01
Rewrite diff check algorithm + improve performance + fix linter fixes 

aminya:
### Description of the change
This updates the way messages are filtered. The previous code was not flagging messages as added, removed, or kept correctly. A highly optimized algorithm is written to do the diff checking.

It greatly improves the performance of linter by not sending messages that are already rendered to linter-ui. 

Fixes https://github.com/steelbrain/linter-ui-default/issues/580
Fixes https://github.com/steelbrain/linter/issues/1709

### should be merged with https://github.com/steelbrain/linter-ui-default/pull/581

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -120,3 +120,13 @@ export function updateKeys(messages: Array<Message>) {
   }
 }
 
+// create a map from keys to messages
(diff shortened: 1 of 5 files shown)


aminya:
The only problem that is remaining is that the actual linter data tips are not removed when the linter message is removed. The markers are removed correctly.

PR #1715, opened 2020-11-05
Fix subfolder git detection

utkarshgupta137:
This change affects the "Ignore VCS" option.

## Description

If you add a project folder which is not a git repo but it contains folders which are git repos, then the repository detection doesn't work.

### Steps to Reproduce

1. Create the following tree:
|-- mainFolderWithoutGit
.....|-- someFolderWithGit
..........|-- .git
2. Add mainFolderWithoutGit as a project.
3. Open a git-ignored file from someFolderWithGit

## Current Behaviour:

The "Ignore VCS" option doesn't work, and the file is still linted.

## Expected behavior:

The file should not be linted if the "Ignore VCS" option is enabled.

## Solution:

The repositoryForDirectory function in atom.project is able to detect the repository for any Directory, even if it is in a subfolder.
It is an async function, so some changes have to be made in up the call chain.

Also updated the tests to test it.


utkarshgupta137:
TODO: Add tests

aminya:
@utkarshgupta137 I changed the codebase to TypeScript. Do you want to move the changes to the new files?

utkarshgupta137:
@aminya done.

aminya:
Could you add some description about the bug this is fixing?

utkarshgupta137:
@aminya done.

aminya (commented):
Thanks for the explanation. I will try this and see if everything works as expected.

lib/helpers.ts
```diff
@@ -24,21 +24,18 @@ export function getEditorCursorScopes(textEditor: TextEditor): Array<string> {
...
   if (ignoredVCS) {
-    let repository = null
-    const projectPaths = atom.project.getPaths()
-    for (let i = 0, { length } = projectPaths; i < length; ++i) {
-      const projectPath = projectPaths[i]
-      if (filePath.indexOf(projectPath) === 0) {
-        repository = atom.project.getRepositories()[i]
-        break
-      }
-    }
```
aminya, line 41: Don't we need to check the project paths anymore?

utkarshgupta137: No, those cases are already covered by `atom.project.repositoryForDirectory`.

aminya (approved):

PR #1722, opened 2021-01-05
feat: allow solutions to be an async callback

aminya:
The linter part of the equation seems to be set!

https://github.com/steelbrain/linter/issues/1705


aminya (commented):
I will merge this for now as it seems there is nothing much to do here. Next, we should fix linter-ui-default

Issue #1712, opened 2020-09-29
User experience pain points from first time user


CubeOfCheese:
Atom version: 1.51.0
OS: Mac OS Catalina 10.15.6
Using with linter-eslint package

Start in a file with no errors:
![Screen Shot 2020-09-28 at 8 31 21 PM](https://user-images.githubusercontent.com/16740469/94509653-9385b600-01c9-11eb-842b-1495d77e2b90.png)
Switch to a file with some errors:
![Screen Shot 2020-09-28 at 8 31 32 PM](https://user-images.githubusercontent.com/16740469/94509663-9a142d80-01c9-11eb-8096-3519f822a4b8.png)
Go back to original file:
![Screen Shot 2020-09-28 at 8 31 41 PM](https://user-images.githubusercontent.com/16740469/94509672-9f717800-01c9-11eb-9790-e4e67e0e0ba5.png)

I expect the number of errors in the linter warnings HUD at the bottom to back down to zero when I switch back to the file with no errors.

Separately, I also expect clicking on the linter warnings HUD to open the linter panel but in the file with no errors it does not. 

Seeing a non zero number of errors in the HUD but not being able to click it to check if there are or are not errors in the current file is confusing. I now understand that if it doesn't open the panel, it means there are no errors in the current file. I also now understand that the number in the HUD is the total for the project rather than the total for the file. I would prefer to see both per project and per file.

PS Thank you to everyone who has worked on this tool!

aminya:
Could you post the screenshots of your linter and linter-ui-default settings. Things you are describing can be customized in the settings.

CubeOfCheese:
Sorry but I'm not sure how to access the settings you're referring to. These are the settings I see:
![Screen Shot 2020-09-28 at 10 06 32 PM](https://user-images.githubusercontent.com/16740469/94514757-def29100-01d6-11eb-9fd4-d07af6c1955e.png)

aminya:
Search for linter-ui-default package

CubeOfCheese:
Thank you! I didn't even realize that package was installed with the other.
![Screen Shot 2020-09-28 at 10 32 16 PM](https://user-images.githubusercontent.com/16740469/94516357-7dccbc80-01da-11eb-84c2-1656fa418e64.png)

CubeOfCheese:
Thank you again. I was able to change the settings to better fit my desired behavior.

[...]

PR #1730, opened 2021-04-14
chore: mention linter-ui-default dependency

mmccarthy15:
Add instructions in the README to prompt user to install the linter-ui-default package if it's not auto-installed.

As a newbie Atom user, I went ahead and installed the linter and code package like all the docs and guides stated.  It took longer than I'd like to admit to find closed Issue #1712 which indicated that the UI package needed to be installed too.

Hope this little addition helps someone else down the line.  Of course, feel free to tweak as you see fit - or let me know if this is something that shouldn't have arisen.  Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ $ apm install linter
 
 Or you can install from Settings view by searching for `Base Linter`, (this package might not show up when searching for `Linter`).
 
+If nothing seems to happen after installing the linter package and at least one language package, you may have to install the `linter-ui-default package`.  This package handles displaying results and errors in the Atom UI.
+
 #### API Documentation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,6 +24,8 @@ $ apm install linter
 
 Or you can install from Settings view by searching for `Base Linter`, (this package might not show up when searching for `Linter`).
 
+If nothing seems to happen after installing the linter package and at least one language package, you may have to install the `linter-ui-default package`.  This package handles displaying results and errors in the Atom UI.
```
aminya, line 27:
```suggestion
Linter automatically installs `linter-ui-default`, which is required for the functionality of this package.
```

aminya (changes requested):
The bug for package-deps is already fixed. We can mention this in the readme so people can report the errors. I prefer not to give them the solution right away, so we know if there is a bug.

